PR #11, opened 2015-02-03
V2.0

tj:
cc @gjohnson & @yields 

Good luck reading the diff haha, but LMK if you see anything weird. You can't really easily pass your own context right now (just manip the default), but meh. We'd have to move the Context field to each struct to make that less stupid


yields:
nice, lgtm! :)

tj:
I dig that, maybe a .type() as well so we can just have a .Send() or just pass them all to the queue method

gjohnson:
Nice, works for me.

PR #16, opened 2015-05-04
Retry and backoff

f2prateek:


analytics.go
```diff
@@ -232,11 +236,24 @@ func (c *Client) send(msgs []interface{}) {
 		return
 	}
 
+	for i := 0; i < 10; i++ {
```
f2prateek: 10 seems like a reasonable default? Could potentially retry it indefinitely as well.

analytics.go
```diff
@@ -283,13 +302,13 @@ func (c *Client) loop() {
 			msgs = append(msgs, msg)
 			if len(msgs) == c.Size {
 				c.verbose("exceeded %d messages – flushing", c.Size)
-				c.send(msgs)
+				go c.send(msgs)
```
f2prateek: Do we need a bound on the number of goroutines this can create?

f2prateek:
cc @tj @gjohnson

analytics.go
tj: Hmm yeah might not be a bad idea, though hopefully we're never down enough for it to back up like mad with retries haha. Maybe just a second goro for handling retries with a high watermark that just drops after, not sure

f2prateek: @tj Very rough draft of what a bounded executor could look like - https://github.com/segmentio/analytics-go/commit/db3adc6760502971fa7356219fd2ba857900f20e

PR #17, opened 2015-05-18
Client.send() should not attempt to make an API if no messages

parkr:
If no messages are passed to Client.send(), then skip the message call altogether.

Wasn't sure how to test this. /cc @tj


tj:
sgtm!

parkr:
thanks @tj!!!

PR #19, opened 2015-07-15
Allow custom timestamps

### LINKED ISSUE #13: Timestamp is always overwritten
`setTimestamp()` is always called upon `queue()`.. disregarding if we had previously set some value there.

Is it normal it uses `time.Now()` and not `time.Now().UTC()` ? so you need to make sure your system is configured with the same timezone as your integrations and all ?

f2prateek:
Closes #13

cc @tj @calvinfo


sperand-io:
thanks @f2prateek!

PR #28, opened 2015-10-14
Fix: Close() method to wait until all messages was queued and sent

vladiacobsendgrid:
Example which is not working for now:

```
client.Interval = 5 * time.Second
client.Size = 51
client.Verbose = true

for i := 0; i < 50; i++ {
	client.Track(&analytics.Track{
		Event:  "TestKeeen2",
		UserId: "125",
		Properties: map[string]interface{}{
			"application": "Segment Desktop",
			"version":     "1.1.0",
			"platform":    "osx",
		},
	})
	println(i)
}

println("flushing")
client.Close()
```

Before that fix, when ``Close()`` method is call, only a few messages are sending to Segment.io. That is happening because these messages was added in c.msgs channel but ``loop()`` didn't had enough time to process and add it in ``var msgs []interface{}``

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -309,10 +309,24 @@ func (c *Client) loop() {
 			}
 		case <-c.quit:
 			c.verbose("exit requested – flushing %d", len(msgs))
+			msgs = append(msgs, c.drain(c.msgs))
 			c.send(msgs)
 			c.verbose("exit")
 			c.quit <- true
 			return
+		}
+	}
+}
+
+func (c *Client) drain(ch chan interface{}) []interface{} {
+	msgs := make([]interface{}, 0)
+	for {
+		select {
+		case msg := <-ch:
+			c.verbose("buffer (%d/%d) %v", len(msgs), c.Size, msg)
+			msgs = append(msgs, msg)
+		default:
+			return msgs
 		}
 	}
 }
--- a/analytics.go
+++ b/analytics.go
@@ -222,8 +222,8 @@ func (c *Client) queue(msg message) {
 // Close and flush metrics.
 func (c *Client) Close() error {
 	c.quit <- true
-	close(c.msgs)
 	<-c.quit
+	close(c.msgs)
 	return nil
 }
 


f2prateek:
This isn't thread safe.

Also I think the logic might be off. It only checks whether a message was queued, not whether it was actually sent, so you're back to square one.

vladiacobsendgrid:
@f2prateek Do you have any other suggestion or can you help me with another suggestion ? We (SendGrid) are trying to implement with you and we really need to wait on close until all messages was sent. 

How I implemented we are waiting until all messages was sent. I'm sending true on quit channel after all messages was added in msgs array so on close all messages queued will be sent and from what I see send() method is synchronous.

f2prateek:
I think what we have is the right behaviour. When you call `Close`, you shouldn't be sending any messages after that, and so it's ok for us to not have to worry about this case.

Just call track/identify/... for messages before you call close.

calvinfo:
@f2prateek I actually ran into this bug as well. The trouble is when you send a bunch of items at once, there's a request currently in-flight, and then flush is called. Depending on how the instructions are interleaved, there's a possible race condition so that messages can be dropped. `Close()` should really wait to close all messages if possible, and drain existing requests.

vladiacobsendgrid:
@f2prateek if you take a look over my example from the PR description you can see I'm sending all the events before to call Close().

f2prateek:
What about something like this, basically drains the msgs queue before uploading in the quit method.

```
diff --git a/analytics.go b/analytics.go
index e013442..af765e9 100644
--- a/analytics.go
+++ b/analytics.go
@@ -309,6 +309,7 @@ func (c *Client) loop() {
                        }
                case <-c.quit:
                        c.verbose("exit requested – flushing %d", len(msgs))
+                       msgs = append(msgs, drain(c.msgs))
                        c.send(msgs)
                        c.verbose("exit")
                        c.quit <- true
@@ -317,6 +318,18 @@ func (c *Client) loop() {
        }
 }

+func drain(c chan interface{}) []interface{} {
+       msgs := make([]interface{}, 0)
+       for {
+               select {
+               case m := <-c:
+                       msgs = append(msgs, m)
+               default:
+                       return msgs
+               }
+       }
+}
+
```

calvinfo:
That will be close. But I think there's still the problem is that the main goro can still exit first without ever waiting for the loop to quit?

vladiacobsendgrid:
@f2prateek I modified my PR to use your suggestions. Can you take a look ? We also tested and is working how we expected.

f2prateek:
You should also move the line here https://github.com/vladiacobsendgrid/analytics-go/blob/master/analytics.go#L225 to after `<-c.quit`, this way the msgs channel isn't closed until it is drained.

vladiacobsendgrid:
@f2prateek done, thanks for suggestions

vladiacobsendgrid:
@f2prateek Can you merge my PR ?

f2prateek:
Thanks!

PR #30, opened 2015-10-20
Clean shutdown

f2prateek:
Adds a second channel so that the background go-routine can indicate
when it finishes uploading events.

Also adds a test for the close method (not for the exact issue, since simply verifies that close flushes the queue).

Behaviour:
* Main goroutine sends a message on the `quit` channel
* Main goroutine closes the `msgs` channel so no more values can be sent
* Main goroutine waits on the `shutdown` channel
* Background goroutine gets the message on the `quit` channel
* Background goroutine drains the `msgs` channel
* Background goroutine uploads messages
* Background goroutine sends a message on the `shutdown` channel
* Main goroutine exits

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -114,7 +114,8 @@ type Client struct {
 	Client   http.Client
 	key      string
 	msgs     chan interface{}
-	quit     chan bool
+	quit     chan struct{}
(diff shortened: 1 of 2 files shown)


calvinfo:
lgtm!

PR #34, opened 2015-10-22
Add Integrations field

ernesto-jimenez:
@f2prateek I have added `Integrations` to each struct rather than `Message` to be consistent with `Context`.

Also, kept it out of `Alias` because `Context` is skipped too, although both can be sent in `alias` messages.


f2prateek:
Thanks!

PR #41, opened 2015-12-26
Defer starting loop.

### LINKED ISSUE #36: Interval is not configurable
Although the interval is public and mutable, since we start the `loop` goroutine right away, clients can't actually change the interval.

f2prateek:
By deferring starting the background loop goroutine, we allow clients to
set an arbitrary flush interval. Previously it would be ignored.

Note: The flush interval can only be configured before any messages are
enqueued.

Closes #36

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -3,6 +3,7 @@ package analytics
 import (
 	"io/ioutil"
 	"os"
+	"sync"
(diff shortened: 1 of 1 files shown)


calvinfo:
:+1:

Issue #39, opened 2015-12-03
Length of messages channel will never exceed Size, and so will never flush


amzuko:
So in analytics.go line 298 we have:

if len(msgs) == c.Size {
...flush...
}

However, msgs is initialized with capacity 100, and Size is initialized with length 250 -- should these two numbers be swapped?

yields:
What version are you using?

f2prateek:
@amzuko I think there is confusion because of the naming of the variable. 

Although the `msgs` channel is [initialized](https://github.com/segmentio/analytics-go/blob/master/analytics.go#L140) with a bound of 100, the line you are [referring to](https://github.com/segmentio/analytics-go/blob/master/analytics.go#L321) is comparing this [local `msgs` array](https://github.com/segmentio/analytics-go/blob/master/analytics.go#L313) and not the global `msgs` channel. This local array can easily grow over 250 in size since it's unbounded.

PR #42, opened 2015-12-26
rename msgs to msgsC

f2prateek:
We're using the `msgs` variable in a few places to refer to different things. This renames the global one to `msgsC` to avoid confusion like in https://github.com/segmentio/analytics-go/issues/39.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -119,7 +119,7 @@ type Client struct {
 	Verbose  bool
 	Client   http.Client
 	key      string
-	msgs     chan interface{}
+	msgsC    chan interface{}
 	quit     chan struct{}
 	shutdown chan struct{}
 	uid      func() string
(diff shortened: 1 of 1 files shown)


f2prateek:
On second though, maybe not since the global on is `c.msgs` instead of just `msgs`.

calvinfo:
Agreed, personally I'm in favor of just keeping it as is

f2prateek:
I still think we should name it `msgC`. To me, `msgs` to represents  a collection, while `msgC` clearly defines a channel by convention from the Go standard lib.

calvinfo:
Doesn't matter too much to me, though I'd say I'm -0 on `msgC`. Most of the stuff outside of lib that I've worked with or read through would be `msgChan` `c` or `msgs`–since the types are there I've found hinting around them to be a bit redundant.

Seems like a small point though so I'm down with whatever.

f2prateek:
cool let's just punt on it for now and if it comes up again we can always re-evaluate.

PR #44, opened 2016-03-25
fixing close

### LINKED ISSUE #43: Calling close before sending anything blocks forever
Hi, thanks for releasing this tool!

I had a weird issue when testing it, where it would block forever if you created a new client, and closed it without sending a single event (in my case a track event):

```golang
func main() {
    client := analytics.New("<segmentWriteKey>")
    client.Close()
}
```

calvinfo:
Fixes #43 

@f2prateek

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -143,7 +143,7 @@ func New(key string) *Client {
 		now:      time.Now,
 		uid:      uid,
 	}
-	c.startLoop()
+
 	return c
 }
 
@@ -225,6 +225,7 @@ func (c *Client) startLoop() {
 
 // Queue message.
 func (c *Client) queue(msg message) {
+	c.once.Do(c.startLoop)
 	msg.setMessageId(c.uid())
 	msg.setTimestamp(timestamp(c.now()))
 	c.msgs <- msg
@@ -232,6 +233,7 @@ func (c *Client) queue(msg message) {
 
 // Close and flush metrics.
 func (c *Client) Close() error {
+	c.once.Do(c.startLoop)
 	c.quit <- struct{}{}
 	close(c.msgs)
 	<-c.shutdown
--- a/analytics.go
+++ b/analytics.go
@@ -143,7 +143,7 @@ func New(key string) *Client {
 		now:      time.Now,
 		uid:      uid,
 	}
-
+	c.startLoop()
 	return c
 }
 
@@ -225,7 +225,6 @@ func (c *Client) startLoop() {
 
 // Queue message.
 func (c *Client) queue(msg message) {
-	c.once.Do(c.startLoop)
 	msg.setMessageId(c.uid())
 	msg.setTimestamp(timestamp(c.now()))
 	c.msgs <- msg
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -3,6 +3,7 @@ package analytics
 import "net/http/httptest"
 import "encoding/json"
 import "net/http"
+import "testing"
 import "bytes"
 import "time"
 import "fmt"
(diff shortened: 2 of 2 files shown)


f2prateek:
The reason we deferred starting the loop was because of this https://github.com/segmentio/analytics-go/issues/36. This change would bring back back the old behaviour of being unable to set a custom flush interval.

f2prateek:
Easier fix for now might be to add c.once.do.c.startloop in the close function as well, ensuring that the loop does start. If it's already been started (by enqueieng a message) it won't do anything and just close the client as normal.

calvinfo:
Hmm, my tests create a race condition when doing that. Any reason we don't
allow options to be passed in or instead do a setter for the interval? It
feels a little weird to me to change an interval on the fly and expect it
to work arbitrarily.

f2prateek:
> Any reason we don't allow options to be passed in or instead do a setter for the interval? 

Mainly coz we're stuck with the legacy API. [Me and @yields](https://github.com/segmentio/analytics-go/issues/36#issuecomment-165873197) have wanted to use a config based approach instead of mutable fields but it's hard to do so without breaking some of the existing functionality/expectations.

paddie:
Have you considered taking a variable list argument of configuration functions?

the `New(writekey string) *Client` would become something like 

```go
type ClientConfigFunc func(*Client) error

New(writekey string, configs ... ClientConfigFunc) (*Client, error) {
    ...
    for _, conf := range configs {
        if err := conf(client); err != nil {
            return nil, err
        }
    }
    ....
}
```

If the for-loop is placed before firing off the loop, there wouldn't be a race, and you could make all the attributes private to people using the client.

paddie:
I should add, that using the variable list arguments, one can completely leave out configurations, and it will still work.

analytics.go
```diff
@@ -143,7 +143,7 @@ func New(key string) *Client {
 		now:      time.Now,
 		uid:      uid,
 	}
-
+	c.startLoop()
```
achille-roussel: Good practices when writing Go code is to have the zero-value of a type be a valid state. In our case it's kind of hard to do because we need at least the write-key (although if not provided we could have code to pick it up from the environment or other places).

So code like this could work:
```go
client := analytics.Client{
    Key: "write-key",
}

// Now we can start using the client
```
If `startLoop` is done in the constructor we force using the constructor to create valid client objects, I believe the current implementation which lazy-loads the goroutine is more powerful.

I see the value in knowing that the goroutine was started when the constructor is called (especially for testing purposes), why not simply use `c.once.Do(c.startLoop)` here instead and not remove it from `(*Client).queue` then? That way we get the best of both approaches?

achille-roussel: Actually based on all the discussions we had today and the directions we're going I don't think this comment is relevant anymore.

achille-roussel:
@calvinfo I think the `once` field isn't required anymore due to this change. Can we remove it and merge?

calvinfo:
Yes, go for it!

achille-roussel:
@f2prateek 
@calvinfo 

Hey guys, I've made the change to call `c.once.Do(c.startLoop)` in the Close method, this is the simplest (maybe only) way to fix the issue while maintaining backward compatibility of the API.

I think we should make a 2.1.1 release today with this bug fix (and #50 that we just merged), these are non-breaking changes, just bug fixes. Then we can make the 3.0.0 release once we've pulled together all the remaining issues that required more refactoring (probably tomorrow).

Let me know what you think of this plan.

f2prateek:
SGTM

calvinfo:
All sounds good to me.

PR #45, opened 2016-03-30
Set message ID only if not present

vinceprignano:
@f2prateek

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -369,7 +369,9 @@ func (m *Message) setTimestamp(s string) {
 
 // Set message id.
 func (m *Message) setMessageId(s string) {
-	m.MessageId = s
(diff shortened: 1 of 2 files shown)


analytics.go
```diff
@@ -369,7 +369,9 @@ func (m *Message) setTimestamp(s string) {
 
 // Set message id.
 func (m *Message) setMessageId(s string) {
-	m.MessageId = s
+	if m.MessageId == "" {
```
calvinfo: I think we should probably change this outside the `setMessageId` method eh? otherwise the name seems a bit misleading

vinceprignano: Good point

f2prateek:
LGTM

PR #49, opened 2016-04-18
Introduce ratelimiter using tokenbucket

paddie:
Introduce ratelimiting of requests to a default (and configurable) rate of 50 per second using the tokenbucket implementation from Canonical's juju.

__NOTE__
I'm basing this PR on the #48 pull-request, and I'll make sure to keep them in sync pending code review etc. So, this should not be merged just yet or at all. Discuss.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/jehiah/go-strftime"
+	"github.com/juju/ratelimit"
 	"github.com/segmentio/backo-go"
 	"github.com/xtgo/uuid"
 )
@@ -25,6 +26,8 @@ const (
 	DefaultBufferSize = 250
 	// Default flush interval
 	DefaultInterval = 5 * time.Second
+	// Default ratelimit is 50 batched requests per second
+	DefaultRatelimit = 50
 )
 
 // DefaultContext of message batches.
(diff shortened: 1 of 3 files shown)


f2prateek:
I'm not sure if we need a dependency for this. Could we simply use `time.NewTicker`?

paddie:
as in, a channel with a requestPerSecond buffer, and have the ticker deposit up to the buffersize to the channel each second?

f2prateek:
Yes, something like.

```go
limiter := time.NewTicker(time.Second/50) // 50rps

// https://github.com/segmentio/analytics-go/blob/master/analytics.go#L271
<-limiter.C
req, err := http.NewRequest("POST", url, bytes.NewReader(b))
```

paddie:
Aye, this would force requests to be spread out over a 1/50 second interval. An alternate approach could be something like this, that would fill up the buffer every second:

```go
tokenChan := make(chan struct{}, c.BufferSize)
for range time.NewTicker(time.Second).C {
    FillerUp()
}

func FillerUp() {
    for {
        select {
        case: tokenChan <- struct{}{}:
        default:
            return
        }
    }
}
```

gjohnson:
Seems out of scope IMO, more of something internal to Segment's API.

achille-roussel:
Yeah I'd rather not bake this into the library so we keep it flexible and don't force rate limiting on the client side if the API can handle more.

@paddie if you don't mind I'll close this pull request, feel free to reopen if you think this does need to be addressed.

Thanks for your contribution!

PR #50, opened 2016-04-19
Reduce blocking when sending

### LINKED ISSUE #72: Client blocks for the duration of the batch marshal and post (including retries)
When the loop receives a message and `len(msgs) == c.size` the list of messages are marshalled into json and sent. When doing this, the client will accept messages up until the buffer channel buffer size, and the block. Worst case, this will be: 
```
jsonMarshalTime + 10 * http-roundtrip + ExponentialRetries
```

This could be a potentially long time (in CPU terms), all the while the client would be unresponsive or just blocking other calls to client.Track etc.

Since whatever errors or complications happened during the send does not really result in anything but a call to log, it might make sense to do this completely async/concurrently.

paddie:
This addresses issue #72 by calling send async, and maintaining an inflight counter and introducing a new error logging case in the main select.

Let me know what you think.

Fixes #72

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -317,7 +317,7 @@ func (c *Client) upload(b []byte) error {
 		return fmt.Errorf("error reading response body: %s", err)
 	}
 
-	return fmt.Errorf("response %s: %s – %s", res.Status, res.StatusCode, string(body))
+	return fmt.Errorf("response %s: %d – %s", res.Status, res.StatusCode, string(body))
 }
 
 // Batch loop.
--- a/analytics.go
+++ b/analytics.go
@@ -126,8 +126,7 @@ type Client struct {
 	uid      func() string
 	now      func() time.Time
 	once     sync.Once
-	inflight uint64
-	sendDone chan error
+	wg       sync.WaitGroup
 }
 
 // New client with write key.
@@ -143,7 +142,6 @@ func New(key string) *Client {
 		msgs:     make(chan interface{}, 100),
 		quit:     make(chan struct{}),
 		shutdown: make(chan struct{}),
-		sendDone: make(chan error),
 		now:      time.Now,
 		uid:      uid,
 	}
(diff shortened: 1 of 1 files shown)


paddie:
@achille-roussel @f2prateek hey guys, could someone give me a CR on this one? It focuses primarily on performance and memory efficiency.

analytics.go
```diff
@@ -239,10 +243,28 @@ func (c *Client) Close() error {
...
+func (c *Client) awaitInflight() {
+	for c.inflight > 0 {
+		c.verbose("waiting for %d requests to finish..", c.inflight)
+		err := <-c.sendDone
+		if err != nil {
+			c.logf(err.Error())
+		}
+		c.inflight--
+	}
+}
```
achille-roussel: This really feels like you're reinventing the wheel, why not simply use a `sync.WaitGroup` object instead? The only benefit I'm seeing is with the code you submitted you get to know how many requests are pending, it doesn't bring a lot of value (it's only ever shown in the logs in verbose mode).

I feel like the right approach is to use a standard library construct, until we have a problem to solve that really needs us to know how many requests are pending (even then we could use a pair of a sync.WaitGroup and an atomic counter to get the information).

paddie: Aye, I see your point. Let me see if I can get the waitgroup stuff in before I have to run out the door.

paddie: @achille-roussel let me know what you think

analytics.go
```diff
@@ -239,10 +241,21 @@ func (c *Client) Close() error {
 	return nil
 }
 
+func (c *Client) sendAsync(msgs []interface{}) {
+	c.wg.Add(1)
+	go func(wg *sync.WaitGroup) {
```
achille-roussel: You don't need to pass it here, since it's already capturing the `c` variable, you can just access `c.wg` in this closure.

paddie:
Good catch, I was all concerned about passing the wg by value that I forgot I had a pointer. fixup'd!

achille-roussel:
The changes look good to me. If you don't mind we'll merge on Monday just in case it has unexpected consequences and break someone's code somewhere. We'd rather not have to fire fight over the weekend.
Thanks for your contribution.

paddie:
No rush, thanks for the CR @achille-roussel !

PR #62, opened 2016-04-21
Close client twice

### LINKED ISSUE #55: Closing a client twice will block
@f2prateek 
@calvinfo 

Currently the `Close` method is not idempotent, and worst it blocks the program's execution if called more than once, see:
```go
// Close and flush metrics.
func (c *Client) Close() error {
    c.quit <- struct{}{}
    close(c.msgs)
    <-c.shutdown
    return nil
}
```
The issue is that nothing is reading from the `c.quit` channel anymore once the client has been closed.

One way of solving this is to close the `c.quit` channel from the client's goroutine before it terminates, but this would cause writing to the channel to panic on the second call of `Close`, so we have to protect this with a recover call, basically changing the method to something like this:
```go
// Close and flush metrics.
func (c *Client) Close() error {
    defer func() {
        // Always recover, a panic could be raised if c.quit was closed which means
        // the Close method was called more than once.
        recover()
    }()
    c.quit <- struct{}{}
    close(c.msgs)
    <-c.shutdown
    return nil
}
```
What do you guys think about it?

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #55 

I've also pulled the changes from #44 in so the test would pass.

Please take a look and let me know if something should be changed!

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -1,9 +1,9 @@
 package analytics
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
-	"sync"
 
 	"bytes"
 	"encoding/json"
@@ -124,7 +124,6 @@ type Client struct {
 	shutdown chan struct{}
 	uid      func() string
 	now      func() time.Time
-	once     sync.Once
 }
 
 // New client with write key.
@@ -143,7 +142,7 @@ func New(key string) *Client {
 		now:      time.Now,
 		uid:      uid,
 	}
-
+	go c.loop()
 	return c
 }
 
@@ -159,7 +158,6 @@ func (c *Client) Alias(msg Alias) error {
 
 	msg.Type = "alias"
 	c.queue(&msg)
-
 	return nil
 }
 
@@ -171,7 +169,6 @@ func (c *Client) Page(msg Page) error {
 
 	msg.Type = "page"
 	c.queue(&msg)
-
 	return nil
 }
 
@@ -187,7 +184,6 @@ func (c *Client) Group(msg Group) error {
 
 	msg.Type = "group"
 	c.queue(&msg)
-
 	return nil
 }
 
@@ -199,7 +195,6 @@ func (c *Client) Identify(msg Identify) error {
 
 	msg.Type = "identify"
 	c.queue(&msg)
-
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


analytics.go
```diff
@@ -310,8 +305,11 @@ func (c *Client) report(res *http.Response) {
 
 // Batch loop.
 func (c *Client) loop() {
+	defer close(c.shutdown)
+
 	var msgs []interface{}
-	tick := time.NewTicker(c.Interval)
+	var tick = time.NewTicker(c.Interval)
+	defer tick.Stop()
```
achille-roussel: This is just a safety measure, the current code flow didn't look like it had an issue but it always safer to clean resources in defer functions... just in case we things start to go wrong we don't happen to leak resources on top of that.

analytics.go
```diff
@@ -215,28 +210,28 @@ func (c *Client) Track(msg *Track) error {
...
-	c.once.Do(c.startLoop)
 	msg.setMessageId(c.uid())
 	msg.setTimestamp(timestamp(c.now()))
 	c.msgs <- msg
 }
 
 // Close and flush metrics.
-func (c *Client) Close() error {
-	c.quit <- struct{}{}
-	close(c.msgs)
```
achille-roussel: I moved this to the client's goroutine, it makes the `Close` method simpler and it's easier to understand where the panic could come from.

analytics.go
```diff
@@ -143,7 +142,7 @@ func New(key string) *Client {
 		now:      time.Now,
 		uid:      uid,
 	}
-
+	go c.loop()
```
f2prateek: Ref: https://github.com/segmentio/analytics-go/pull/41

My moving this here, we essentially don't give users a way to configure https://github.com/segmentio/analytics-go/blob/close-client-twice/analytics.go#L311.

Once we have the config stuff, though, this should be fine 👍

achille-roussel: You're right, I was thinking of merging this into the `v3.0` branch first and not master so it shouldn't break anything (same for all the stuff we discussed today).

analytics.go
```diff
@@ -332,17 +330,25 @@ func (c *Client) loop() {
 				c.verbose("interval reached – nothing to send")
 			}
 		case <-c.quit:
-			tick.Stop()
 			c.verbose("exit requested – draining msgs")
-			// drain the msg channel.
+
```
calvinfo: This is a good point–any reason not to just set a flag on the client like `.closing` and check that pre-enqueue?

calvinfo: Agreed that it shouldn't block this PR

analytics.go
calvinfo: nice, good catch

analytics.go
calvinfo: It's also weirded me out that any of our libs would have an interval. It feels like it makes more sense to just limit the number of outstanding requests (since that is the desired bottleneck) and then take messages off up to `max_batch` at a time. 

Thoughts? That feels the most natural to me anyway, just having a consumer constantly running, and taking messages off in batches of up to 100.

calvinfo: Basically, I've never seen the *need* for an interval in any of our libraries, just a max-in-flight number of requests, that we would set at 1 by default.

analytics.go
achille-roussel: The flag would need to be atomic or protected by a lock, and there's always the risk of the state of that flag and the 'close' state of the channel to get out of sync if someone makes a change and didn't anticipate the side effects.
I feel like the channel here gives us all we need already, we would only need to recover from the panic in the enqueue method and return an error when it's already closed, plus it doesn't slow down the normal flow with extra synchronization work. 

On a side note we *have to* close the channel anyway to use a simple for loop to flush it.

analytics.go
achille-roussel: I think this was mostly interesting when doing development, if the in-flight message limit is high then the tests likely won't post the events until the program is stopped (which usually happens with a hard kill in dev so the events aren't sent at all). 
I agree that setting the limit low seems like a simpler way of solving this problem.

Issue #53, opened 2016-04-20
Merging all send methods into one


achille-roussel:
@f2prateek 

I think we could make the client API simpler if we were merging all send methods into one `Send(interface{})` and doing a type-check internally to know what message has been passed, that we we could write things like:
```go
client.Send(analytics.Track{ ... })
client.Send(analytics.Page{ ... })
...
```
which is less redundant than
```go
client.Track(analytics.Track{ ... })
client.Page(analytics.Page{ ... })
```
and also makes the API simpler, therefore easier to document and learn?

The downside I'm seeing is that we lose the compile-time check that the program is passing valid types to the function, although we can get it back with a trick like this:
```go
type Event interface {
    // This method is unexported so only types defined in the analytics package can
    // implement this interface.
    path() string
}

type Alias struct { ... }

func (a Alias) path() string { return "/alias" }

// Expects a type that satisfies the Event interface
func (c *Client) Send(e Event) error { ... }
```

f2prateek:
Yup I think a single enqueue would be better (as we do in the java library) https://segment.com/docs/libraries/java/.

achille-roussel:
Cool! Yeah let's keep the naming as consistent across libraries and call it `Enqueue` then.

f2prateek:
Cool, yeah I preferred `enqueue` in the Java library over `send` since it's clearer that the message is not uploaded right away.

PR #63, opened 2016-04-21
pass parameters by value to client methods

### LINKED ISSUE #52: Pass values instead of pointers to send methods
@f2prateek 

What do you think about passing values instead of pointers to the send methods? Besides making the syntax heavier and putting more pressure on the compiler and GC I don't really see why this is done (we're not using the `nil` value for example and it would always cause a panic to pass `nil`, crashing at runtime instead of compile time).

Basically changing
```go
func (c *Client) Alias(a *Alias) error { ... }
```
to
```go
func (c *Client) Alias(a Alias) error { ... }
```
Let me know what you think about it?

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #52 

This is an intermediary change toward #53, please review and let me know if anything should be changed.

### DIFF
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -79,7 +79,7 @@ func ExampleTrack() {
 	//   "context": {
 	//     "library": {
 	//       "name": "analytics-go",
-	//       "version": "2.0.0"
+	//       "version": "2.1.0"
 	//     }
 	//   },
 	//   "messageId": "I'm unique",
@@ -128,7 +128,7 @@ func ExampleClose() {
 	//   "context": {
 	//     "library": {
 	//       "name": "analytics-go",
-	//       "version": "2.0.0"
+	//       "version": "2.1.0"
 	//     }
 	//   },
 	//   "messageId": "I'm unique",
@@ -176,7 +176,7 @@ func ExampleInterval() {
 	//   "context": {
 	//     "library": {
 	//       "name": "analytics-go",
-	//       "version": "2.0.0"
+	//       "version": "2.1.0"
 	//     }
 	//   },
 	//   "messageId": "I'm unique",
@@ -227,7 +227,7 @@ func ExampleTrackWithTimestampSet() {
 	//   "context": {
 	//     "library": {
 	//       "name": "analytics-go",
-	//       "version": "2.0.0"
+	//       "version": "2.1.0"
 	//     }
 	//   },
 	//   "messageId": "I'm unique",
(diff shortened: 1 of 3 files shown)


achille-roussel:
The Circle trace is strange, it looks like it tested against an old version of the library but with the newer tests:
```
# _/home/ubuntu/analytics-go/examples
examples/track.go:22: cannot use analytics.Track literal (type analytics.Track) as type *analytics.Track in argument to client.Track
ok  	_/home/ubuntu/analytics-go	5.013s	coverage: 62.9% of statements
godep: go exit status 2
make: *** [test] Error 1

```

achille-roussel:
Definitely passes when I run it:
```
$ make test
ok  	github.com/segmentio/analytics-go	5.020s	coverage: 61.2% of statements
?   	github.com/segmentio/analytics-go/examples	[no test files]
Achilles-MacBook-Pro:analytics-go achilleroussel$ git status
On branch pass-by-value
Your branch is up-to-date with 'origin/pass-by-value'.
nothing to commit, working directory clean
```

f2prateek:
Yup the thing with circle is that it fetches master rather than the Github src directory. We should be moving the directory into the gopath.

achille-roussel:
@f2prateek alright it looks like the tests are passing now! I had to make changes to the tests so that they would pass `go vet`, basically I converted a bunch of examples to actual unit tests (it should have no impact since they didn't even show up properly in the doc).

Please take a look and let me know if you're good with the changes.

analytics_test.go
```diff
@@ -1,6 +1,9 @@
 package analytics
 
-import "net/http/httptest"
```
calvinfo: mind making these all follow the same imports? just for consistency :)

analytics_test.go
```diff
@@ -108,44 +111,51 @@ func ExampleClose() {
...
-	//   "context": {
-	//     "library": {
-	//       "name": "analytics-go",
-	//       "version": "2.1.0"
-	//     }
-	//   },
-	//   "messageId": "I'm unique",
-	//   "sentAt": "2009-11-10T23:00:00+0000"
-	// }
+	const ref = `{
```
calvinfo: ❤️

calvinfo:
LGTM

analytics_test.go
achille-roussel: Hm... Not sure why gofmt didn't fix this, I'll do!

achille-roussel:
Thanks for reviewing, I had based this PR on master which was a mistake, see #65 which merges those changes into `v3.0`.

Closing.

PR #64, opened 2016-04-21
Update Circle Config

f2prateek:
Makefile: Enable race detector, fix phony targets.

Circle: Clone the repo into our Gopath so circle doesn't fetch master
but actually uses the code in commit.

Go Vet fixes:
1. rename log to logf so go vet isn't confused
2. use %d for response.StatusCode instead of %s

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,7 +12,7 @@ dependencies:
 
 test:
   pre:
-    - go vet ./...
+    - make vet
 
   override:
     - make test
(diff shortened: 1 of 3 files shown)


f2prateek:
@achille-roussel margin this should fix the CI issues in #63

circle.yml
```diff
@@ -1,8 +1,18 @@
...
+  pre:
     - go get github.com/tools/godep
 
+  override:
+    - mkdir -p "$GOPATH/src/$IMPORT_PATH"
+    - rsync -azC --delete ./ "$GOPATH/src/$IMPORT_PATH/"
+
 test:
+  pre:
+    - go vet ./...
```
achille-roussel: Is there a reason not to make this a `make` rule?

f2prateek: Good idea, done!

achille-roussel:
Looks good 👍 
Thanks for fixing!

PR #66, opened 2016-04-21
Close client twice (v3.0)

### LINKED ISSUE #55: Closing a client twice will block

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #55 

Follow up of #62 but will merge into `v3.0` instead of `master`.

Please let me know if anything needs to be changed.

### DIFF
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -493,3 +493,15 @@ func TestTrackWithIntegrations(t *testing.T) {
 		t.Errorf("invalid response:\n- expected %s\n- received: %s", ref, res)
 	}
 }
+
+func TestCloseTwice(t *testing.T) {
+	client := New("0123456789")
+
+	if err := client.Close(); err != nil {
+		t.Error("closing a client should not a return an error")
+	}
+
+	if err := client.Close(); err != io.EOF {
+		t.Error("closing a client a second time should return io.EOF:", err)
+	}
+}
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -495,19 +495,33 @@ func TestTrackWithIntegrations(t *testing.T) {
 }
 
 func TestCloseTwice(t *testing.T) {
+	// This test is overly complex for what it attempts to verify but currently
+	// closing the client without sending anythinng blocks, I'll cleanup later.
+	body, server := mockServer()
+	defer server.Close()
+
 	client := New("0123456789")
+	client.Endpoint = server.URL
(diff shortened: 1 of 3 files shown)


analytics_test.go
```diff
@@ -493,3 +493,27 @@ func TestTrackWithIntegrations(t *testing.T) {
...
+	if err := client.Close(); err != nil {
+		t.Error("closing a client should not a return an error")
+	}
+
+	if err := client.Close(); err != io.EOF {
+		t.Error("closing a client a second time should return io.EOF:", err)
+	}
+}
+
+func TestCloseTwice(t *testing.T) {
```
f2prateek: I think we can remove this one.

achille-roussel: Oops, copy/paste error... good catch.

achille-roussel:
@f2prateek 

The tests are still failing with this trace:
```
# _/home/ubuntu/analytics-go/examples
examples/track.go:22: cannot use analytics.Track literal (type analytics.Track) as type *analytics.Track in argument to client.Track
ok  	_/home/ubuntu/analytics-go	1.145s	coverage: 62.2% of statements
godep: go exit status 2
make: *** [test] Error 1

```
Any idea what may be happening there? I do have the latest circle.yml file in this branch so I don't understand what's going on.

achille-roussel:
Looks like it was a caching issue, Calvin helped me figuring it out.

f2prateek:
cool LGTM!

PR #67, opened 2016-04-21
add analytics.Logger interface + .gitignore

### LINKED ISSUE #54: Don't force using a log.Logger
@f2prateek 

Right now we allow configuring the client logger by defining a `*log.Logger` object that a client should use.
This does limit the loggers usable to the logger from the standard Go library but there are other logging libs out there that may be used and that we could integrate with.

We only use the `Printf` method of the logger object we receive, I think we could take one of these two approaches, which comes handy when integrating with unit tests as well:
- using an interface
```go
type Logger interface {
    Logf(string, ...interface{})
}

type Client struct {
    ...
    Logger Logger
}
```
- using a function
```go
type Client struct {
    ...
    Logf func(string, ...interface{})
}
```
I tend to like the first approach better as it's usually how it's done in Go, it also makes it easy to integrate with the unit tests logger:
```go
func TestX(t *testing.T) {
     client := analytics.New("write-key")
     client.Logger = t // cloud also be `client.Logf = t.Logf` with the other option
     ...
}
```
Go makes it also really easy to make functions implement single-method interfaces with tricks like this:
```go
type LoggerFunc func(string, ...interface{})

func (f LoggerFunc) Logf(format string, args ...interface{}) {
    f(format, args...)
}
...
client.Logger = analytics.LoggerFunc(logf)
```
so we could get the best of both worlds this way.

Let me know what you think about it.

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #54 

These changes add the `analytics.Logger` interface to make it easier to plug any logger to instances of `analytics.Client`.

Please take a look and let me know if anything should be changed.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -35,11 +35,11 @@ type stdLogger struct {
 }
 
 func (l stdLogger) Logf(format string, args ...interface{}) {
-	l.logger.Printf("INFO "+format, args...)
+	l.logger.Printf("INFO: "+format, args...)
 }
 
 func (l stdLogger) Errorf(format string, args ...interface{}) {
-	l.logger.Printf("ERROR "+format, args...)
+	l.logger.Printf("ERROR: "+format, args...)
 }
 
 func newDefaultLogger() Logger {
--- a/logger.go
+++ b/logger.go
@@ -1,0 +1,47 @@
+package analytics
+
+import (
+	"log"
+	"os"
+)
+
+// Instances of types implementing this interface can be used to define where
+// the analytics client logs are written.
+type Logger interface {
+
+	// Analytics clients call this method to log regular messages about the
+	// operations they perform.
+	// Messages logged by this method are usually tagged with an `INFO` log
+	// level in common logging libraries.
+	Logf(format string, args ...interface{})
+
+	// Analytics clients call this method to log errors they encounter while
+	// sending events to the backend servers.
+	// Messages logged by this method are usually tagged with an `ERROR` log
+	// level in common logging libraries.
+	Errorf(format string, args ...interface{})
+}
+
+// This function instantiate an object that statisfies the analytics.Logger
+// interface and send logs to standard logger passed as argument.
+func StdLogger(logger *log.Logger) Logger {
+	return stdLogger{
+		logger: logger,
+	}
+}
+
+type stdLogger struct {
+	logger *log.Logger
+}
+
+func (l stdLogger) Logf(format string, args ...interface{}) {
+	l.logger.Printf("INFO "+format, args...)
+}
+
+func (l stdLogger) Errorf(format string, args ...interface{}) {
+	l.logger.Printf("ERROR "+format, args...)
+}
+
+func newDefaultLogger() Logger {
+	return StdLogger(log.New(os.Stderr, "segment ", log.LstdFlags))
+}
--- a/logger_test.go
+++ b/logger_test.go
@@ -25,9 +25,9 @@ func TestStdLogger(t *testing.T) {
 	logger.Logf("The answer is %d", 42)
 	logger.Errorf("%s", errors.New("something went wrong!"))
 
-	const ref = `test INFO Hello World!
-test INFO The answer is 42
-test ERROR something went wrong!
+	const ref = `test INFO: Hello World!
+test INFO: The answer is 42
+test ERROR: something went wrong!
 `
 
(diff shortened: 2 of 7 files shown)


achille-roussel:
Tests are failing because it needs #66 to be merged first.

logger.go
```diff
@@ -0,0 +1,47 @@
...
+		logger: logger,
+	}
+}
+
+type stdLogger struct {
+	logger *log.Logger
+}
+
+func (l stdLogger) Logf(format string, args ...interface{}) {
+	l.logger.Printf("INFO "+format, args...)
```
f2prateek: would this be better as `INFO: ` (same for error)?

achille-roussel: I have no strong opinion on it, I'll change it if you like it better this way.

logger_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (t *testLogger) Logf(format string, args ...interface{}) {
+	t.logs += fmt.Sprintf("INFO "+format+"\n", args...)
+}
+
+func (t *testLogger) Errorf(format string, args ...interface{}) {
+	t.errors += fmt.Sprintf("ERROR "+format+"\n", args...)
+}
+
+// This test ensures that the interface doesn't get changed and stays compatible
```
f2prateek: could we just do `var _ Logger = (*testing.T)(nil)` to assert interface compliance?

achille-roussel: That's a good idea, I went a little overkill with this one :P

achille-roussel:
@f2prateek I made the changes you requested.

achille-roussel:
@f2prateek Circle-CI is being mean again, do you know why it's bugging us with this?
```
analytics.go:15:2: cannot find package "github.com/segmentio/backo-go" in any of:
	/usr/local/go/src/github.com/segmentio/backo-go (from $GOROOT)
	/home/ubuntu/analytics-go/Godeps/_workspace/src/github.com/segmentio/backo-go (from $GOPATH)
	/home/ubuntu/.go_workspace/src/github.com/segmentio/backo-go
	/usr/local/go_workspace/src/github.com/segmentio/backo-go
examples/track.go:3:8: cannot find package "github.com/segmentio/analytics-go" in any of:
	/usr/local/go/src/github.com/segmentio/analytics-go (from $GOROOT)
	/home/ubuntu/analytics-go/Godeps/_workspace/src/github.com/segmentio/analytics-go (from $GOPATH)
	/home/ubuntu/.go_workspace/src/github.com/segmentio/analytics-go
	/usr/local/go_workspace/src/github.com/segmentio/analytics-go
godep: go exit status 1
make: *** [test] Error 1

```
I've tried doing `godep save` (the backo-go dependency was missing), then building without cache, but it didn't help fixing the issue.

achille-roussel:
Alright, it looks like I managed to make it work, I modified the circle.yml file to match what's used in other repos we have.

logger_test.go
```diff
@@ -0,0 +1,36 @@
+package analytics
+
+import (
+	"bytes"
+	"errors"
+	"log"
+	"testing"
+)
+
+// This test ensures that the interface doesn't get changed and stays compatible
```
f2prateek: Not sure this needs to be in a test since there's no point running this (it breaks at compile time).

f2prateek:
LGTM

logger_test.go
f2prateek: cc @achille-roussel

achille-roussel: I think this is the point, we want it to break at compile time to ensure that the interface stays compatible. For example if we introduce a method in the `Logger` interface that doesn't exist in `*testing.T` then we want something to tell us we're probably getting ourselves in trouble.

f2prateek: Correct, I meant instead of 

```go
// This test ensures that the interface doesn't get changed and stays compatible
// with the *testing.T type.
// If someone were to modify the interface in backward incompatible manner this
// test would break.
func TestTestingLogger(t *testing.T) {
	_ = Logger(t)
}
```

we could really just do:

```go
var _ Logger = (*testing.T)(nil) // Assert interface compliance.
```

achille-roussel: Oh I see what you mean, just moving it to the global scope instead of having a test that actually get run?

Yeah that would work.

f2prateek: Yup! Not a big deal though, since either way it'll break when it's not working and and not break when it's working so fine to leave it as is.

PR #70, opened 2016-04-22
Enqueue

### LINKED ISSUE #53: Merging all send methods into one

### LINKED ISSUE #56: Use time.Time to represent timestamps in data structures
@f2prateek 
@calvinfo 

Currently the timestamp and sentAt fields have a `string` type in the data structures (`Alias`, `Track`, ...), this is not very handy and developers have to refer to the HTTP API documentation to understand what format this should be in (which isn't linked from the Go code).

I feel like we should use `time.Time` for these fields and take care of the serialization internally so we take the burden off of the developers using the library and leverage the compiler to verify that they're using valid values (basically a time value, right now they could inject whatever string format in there).

Let me know what you think about it?

### LINKED ISSUE #58: Get rid of the Message type
@f2prateek 

I think we should get rid of the `Message` type and the inheritance-like approach in the current implementation and instead inject the `MessageId`, `SentAt` and `Timestamp` field into each object.

The thing that's bugging me the most is the exported `Type` field and the `message` interface hack. `Type` is only ever used internally and doesn't need to be exposed and we wouldn't even need it internally if we took the approach defined in #53

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #58 
Fixes #56 
Fixes #53 

Hey guys, this pull request fixes the three issues mentioned above and adds a bunch of tests, it's more changes at once that I was expecting to make but as I was digging into the problems I realized making the changes individually would have required a lot of temporary code to be written to make things work.

Please take a look and let me know if you see anything that should be changed.

Side note: code coverage is now at 82% ;) I'm targeting 100% for the next release.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -8,33 +8,35 @@ type Alias struct {
 	MessageId  string
 	PreviousId string
 	UserId     string
-	SentAt     time.Time
 	Timestamp  time.Time
 }
 
-const (
-	alias = "alias"
-)
-
 func (msg Alias) validate() error {
 	if len(msg.UserId) == 0 {
-		return MissingFieldError{Type: alias, Name: "UserId"}
+		return FieldError{
+			Type:  "analytics.Alias",
+			Name:  "UserId",
+			Value: msg.UserId,
+		}
 	}
 
 	if len(msg.PreviousId) == 0 {
-		return MissingFieldError{Type: alias, Name: "PreviousId"}
+		return FieldError{
+			Type:  "analytics.Alias",
+			Name:  "PreviousId",
+			Value: msg.PreviousId,
+		}
 	}
 
 	return nil
 }
 
-func (msg Alias) serializable() interface{} {
+func (msg Alias) serializable(msgid string, time time.Time) interface{} {
 	return serializableAlias{
-		Type:       alias,
-		MessageId:  msg.MessageId,
+		Type:       "alias",
+		MessageId:  makeMessageId(msg.MessageId, msgid),
 		PreviousId: msg.PreviousId,
 		UserId:     msg.UserId,
-		SentAt:     formatTime(msg.SentAt),
 		Timestamp:  formatTime(msg.Timestamp),
 	}
 }
@@ -44,6 +46,5 @@ type serializableAlias struct {
 	MessageId  string `json:"messageId,omitempty"`
 	PreviousId string `json:"previousId"`
 	UserId     string `json:"userId"`
-	SentAt     string `json:"sentAt,omitempty"`
 	Timestamp  string `json:"timestamp,omitempty"`
 }
--- a/alias.go
+++ b/alias.go
@@ -5,10 +5,12 @@ import "time"
 // This type represents object sent in a alias call as described in
 // https://segment.com/docs/libraries/http/#alias
 type Alias struct {
-	MessageId  string
-	PreviousId string
-	UserId     string
-	Timestamp  time.Time
+	MessageId    string
+	PreviousId   string
+	UserId       string
+	Timestamp    time.Time
+	Context      map[string]interface{}
+	Integrations map[string]interface{}
 }
 
 func (msg Alias) validate() error {
@@ -33,18 +35,22 @@ func (msg Alias) validate() error {
 
 func (msg Alias) serializable(msgid string, time time.Time) interface{} {
 	return serializableAlias{
-		Type:       "alias",
-		MessageId:  makeMessageId(msg.MessageId, msgid),
-		PreviousId: msg.PreviousId,
-		UserId:     msg.UserId,
-		Timestamp:  formatTime(msg.Timestamp),
+		Type:         "alias",
+		MessageId:    makeMessageId(msg.MessageId, msgid),
+		PreviousId:   msg.PreviousId,
+		UserId:       msg.UserId,
+		Timestamp:    formatTime(msg.Timestamp),
+		Context:      msg.Context,
+		Integrations: msg.Integrations,
 	}
 }
 
 type serializableAlias struct {
-	Type       string `json:"type,omitempty"`
-	MessageId  string `json:"messageId,omitempty"`
-	PreviousId string `json:"previousId"`
-	UserId     string `json:"userId"`
-	Timestamp  string `json:"timestamp,omitempty"`
+	Type         string                 `json:"type,omitempty"`
+	MessageId    string                 `json:"messageId,omitempty"`
+	PreviousId   string                 `json:"previousId"`
+	UserId       string                 `json:"userId"`
+	Timestamp    string                 `json:"timestamp,omitempty"`
+	Context      map[string]interface{} `json:"context,omitempty"`
+	Integrations map[string]interface{} `json:"integrations,omitempty"`
 }
--- a/alias.go
+++ b/alias.go
@@ -1,0 +1,49 @@
+package analytics
+
+import "time"
+
+// This type represents object sent in a alias call as described in
+// https://segment.com/docs/libraries/http/#alias
+type Alias struct {
+	MessageId  string
+	PreviousId string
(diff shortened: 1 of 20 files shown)


achille-roussel:
The Circle-CI build fails because I need changes from #67

alias.go
```diff
@@ -0,0 +1,50 @@
+package analytics
+
+import "time"
+
+// This type represents object sent in a alias call as described in
+// https://segment.com/docs/libraries/http/#alias
+type Alias struct {
+	MessageId  string
```
f2prateek: missing a few top level properties here? e.g. integrations

f2prateek:
I'm not sure what the advantage of having an internal serializable representation of all the events is. I think we can cut down a lot of code by just making it return a map[string]interface{}

f2prateek:
Or even if we make the types simply implement the `json.Marshaler` interface?

calvinfo:
It does feel a little verbose, was that to be able to rename the fields easily and more flexibly than what the tags give you?

calvinfo:
Otherwise, lgtm!

alias.go
achille-roussel: Indeed, thanks for pointing it out. It was missing in the current version of the library but the docs do mention those fields.

achille-roussel:
- I'd like to avoid having to use maps to represent the messages, using structures gives us type safety and very expressive code that is more self documenting, plus there's always a runtime cost when dealing with dynamic data structures vs something that's defined at compile time.

- I think we can make the types satisfy `json.Marshaler` at a later time and it can be built on top of the implementation proposed here

- it is a little verbose but the API exposed to the consumer of the library is actually smaller than before, we're reducing complexity for the consumer, but the trade off is increasing it a little internally, and there's heavy testing in place to ensure we keep this under control

The problem this was intended to solve is the initial implementation was using unnecessarily complex structures with sudo-inheritance and exposing fields that belonged to the implementation. Here we're trying to simplify the user-facing API.

I'm gonna fix the Circle-CI build, @f2prateek if you feel good about it I'll merge then.

f2prateek:
> I'd like to avoid having to use maps to represent the messages, using structures gives us type safety and very expressive code that is more self documenting,

I'm not sure it really does . Yes, we verify that we're accidentally passing a `string` instead of `int`  for a field internally. But we have tests to verify the Json output so we gain very little for this type safety (since that's the only purpose of the internal type).

>  plus there's always a small runtime cost when dealing with dynamic data structures vs something that's defined at compile time.

Hmm... I think if anything, the current implementation has higher runtime cost. The json marshaller has to read the fields of `serializableAlias` (and the other internal types) using reflection. If we returned a `map[string]interface{}` - the marshaller just reads the keys of the map and not read reflection.

Alternatively by directly implementing the marshaller interface - we can eliminate the reflection lookups/map marshalling entirely.

achille-roussel:
OK I understand what you mean about using maps that way.

Actually even when using maps the encoder is likely to use reflection just as much as it still doesn't know what types the values are in `map[string]interface{}`, the only optimization that could be done is doing a type check for `map[string]interface{}` because it's a common type ([although it doesn't look like that's done](https://golang.org/src/encoding/json/encode.go#L358)), but after that it has to use reflection.

By implementing `json.Marshaler` we could optimize and avoid all use of reflection but I doubt it would be worth the effort, a simple implementation of the method would look like this:
```go
func (msg Type) MarshalJSON() ([]byte, error) {
    return json.Marhsal(/* map or struct goes here */)
}
```
So the call to `json.Marshal` within `MarshalJSON` will then fallback on a reflective approach internally...

f2prateek:
> it is a little verbose but the API exposed to the consumer of the library is actually smaller than before, we're reducing complexity for the consumer, but the trade off is increasing it a little internally, and there's heavy testing in place to ensure we keep this under control.

I think we've increased the internal complexity a **lot**. And we've only reduced a little for the consumer:

1. Hide 2 internal fields `Type` and `SentAt` (that if passed by the would be ignored).
2. Moved `Timestamp`, `MessageId` into top level which makes them more convenient to use.
3. Merge multiple methods (`Group`, etc.) into one Method (`Enqueue`) which IMO is a neutral change.

To me, the reduction external complexity isn't enough to justify the increased internal complexity.

f2prateek:
Maybe I'm hesitant because this PR feels big because it solves a lot at once. I think #56 should definitely be it's own. And I think #58 can be implemented w/o #53?

achille-roussel:
1. I agree, this is not an important change, we could have gone away with it just by providing better documentation, `SendAt` doesn't need to be exposed since it only needs to be set in the `batch` object which we don't need to export. Not exposing the `Type` field in other messages is actually the only driver behind most of the changes here.
2. I think we both agree this is a good change and it also wouldn't have required any implementation changes
3. This also wouldn't require the refactor, it's a minor change that I think brings good value because it makes the `analytics.Client` interface so much simpler (2 methods vs 6, that's less to document and learn).

I think the reason I'm feeling comfortable with the change is because everything is covered by tests that dynamically inspects the structure and verifies that they will be modified properly in the future.

About splitting the changes, it's tricky to make the changes separately without writing a bunch of intermediary code because the existing implementation really didn't fit using `time.Time` fields for example.

I'd be OK putting the `Type` field back in the message structure to get rid of the *serializable-* types if you feel like it would be a better trade off. I'll just document that it's here for serialization purposes and should be ignored by the application.

f2prateek:
> I'd be OK putting the Type field back in the message structure to get rid of the serializable- types if you feel like it would be a better trade off. I'll just document that it's here for serialization purposes and should be ignored by the application.

Cool, I think that would definitely reduce the internal complexity.

Let's merge this PR as is for now (and we can make the improvements in future PRs), so it doesn't block any other pending work.

LGTM :shipit:

achille-roussel:
Although if we don't use the `serializable` method to generate the serializable version of the object anymore there are still some complexity that will just be moved instead of removed, for example setting the message id and timestamp, we'll need something like we had in the current code.
```go
type Message interface {
    ...
    setMesssageId(string)
    setTimestamp(time.Time)
}
```
Maybe we can merge the two into one function like:
```go
type Message interface {
    ...
    setDefaults(msgid string, timestamp time.Time)
}
```
Or take the approach that I think you were suggesting with a map and having an implementation like this:
```go
func (msg Track) serializable(msgid string, timestamp time.Time) map[string]interface{} {
    return map[string]interface{}{
        "type": "track",
        ...
    }
}
```
Note that this does remove the `serializable-` types but we still need just as many tests to verify that the returned maps are matching the structure they were generated from.

I've been through each of these steps during development just to realize that the solution I ended up with was the one I found to be the best trade off between public / private complexity and potential runtime costs, I'm totally opened to hearing your thoughts on those options or even suggest another one.

Also I feel like most of the lines of code I added in the PR are actually unit tests, which is great, the apparent complexity in the code itself is really just showing up in the pair of public and serializable types I think and isn't that much (I mean it's all declarative code that's pretty straight-forward to read).

achille-roussel:
OK cool, thanks for your reviews man!

PR #73, opened 2016-04-23
use http.RoundTripper instead of http.Client

### LINKED ISSUE #46: Redundant Close() called in Client.report(resp) method
c.report is only called in one location, but both the caller and report method closes the Body:

```go
    ...
	res, err := c.Client.Do(req)
	if err != nil {
		c.log("error sending request: %s", err)
		return err
	}
	defer res.Body.Close()

	c.report(res)

	return nil
}

// Report on response body.
func (c *Client) report(res *http.Response) {
	if res.StatusCode < 400 {
		c.verbose("response %s", res.Status)
		return
	}

	defer res.Body.Close() <------ redundant
	body, err := ioutil.ReadAll(res.Body)
	if err != nil {
		c.log("error reading response body: %s", err)
		return
	}

	c.log("response %s: %s – %s", res.Status, res.StatusCode, body)
}
```

https://github.com/segmentio/analytics-go/blob/master/analytics.go#L301

### LINKED ISSUE #57: Use http.RoundTripper instead of http.Client
@f2prateek 

We're currently allowing the `analytics.Client` to configure the HTTP client it uses to connect to our API. While this is great it offers more flexibility than is usually needed.

http.Client is not a generic type and it allows configuring more than we need/want. For example it manages cookies, redirections, timeouts... things that we potentially want to have full control over.

My understanding is that we want to allow overriding the transport layer (let's say to do unit tests that don't actually connect over the network, use a different pooling policy, etc...) but the rest of the setup should be handled by the library.

I think we could change this to have a `Transport http.RoundTripper` field in the `analytics.Client`, which is an interface so consumers of the library can really inject whatever they want here.

Thoughts?

achille-roussel:
@f2prateek 

Fixes #46 
Fixes #57 

Simple changes related to the two issues mentioned above, please let me know if you want to see something changed.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -38,35 +38,35 @@ type Client struct {
 	Endpoint string
 	// Interval represents the duration at which messages are flushed. It may be
 	// configured only before any messages are enqueued.
-	Interval time.Duration
-	Size     int
-	Logger   Logger
-	Verbose  bool
-	Client   http.Client
-	key      string
-	msgs     chan interface{}
-	quit     chan struct{}
-	shutdown chan struct{}
-	uid      func() string
-	now      func() time.Time
-	once     sync.Once
+	Interval  time.Duration
+	Size      int
+	Logger    Logger
+	Verbose   bool
+	Transport http.RoundTripper
+	key       string
+	msgs      chan interface{}
+	quit      chan struct{}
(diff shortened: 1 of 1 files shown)


f2prateek:
LGTM

PR #74, opened 2016-04-23
No godep

### LINKED ISSUE #69: Drop godep dependency
@f2prateek 
@calvinfo 

Hey guys, what do you think about relying on the standard vendoring system (available from Go 1.5+ with `GO15VENDOREXPERIMENT=1` set, and enabled by default in Go 1.6)?

Circle-CI is building with Go 1.6 already according to [this](https://circleci.com/docs/build-image-trusty/).

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #69 

With this PR we no longer depend on Godep for dependency management but instead use the standard approach that's now available in Go 1.6


f2prateek:
LGTM!

PR #77, opened 2016-04-24
Context

### LINKED ISSUE #59: Add the Context type
@f2prateek 

The sepcs says the context field is an object with a specific structure (unlike properties or traits which are free-form objects).

How about defining an actual struct type to make the code more self-documenting on what can be set?

I guess my main question is, could this be too restrictive in some cases or is there any undocumented parts of the context that our API actually supports?

https://segment.com/docs/spec/common/#context

achille-roussel:
@f2prateek 
@calvinfo 

Fix #59 

The changes here add the definitions of the context type an sub-types, please take a look and let me know if you're satisfied with the changes.

Highlights:
- The `Extra` field in the object is used for undocumented extensions, it's flattened when the context is serialized.
- To support flattening the `Extra` field but also to work around the limitations of the standard `json` package (like not considering *empty* any struct type zero-value) the `Context` type satisfies `json.Marshaler` to hijack the default serialization. It builds a map representation of the value and serialize that instead.


analytics.go
```diff
@@ -20,14 +20,6 @@ const Version = "3.0.0"
 // Endpoint for the Segment API.
 const Endpoint = "https://api.segment.io"
 
-// DefaultContext of message batches.
-var DefaultContext = map[string]interface{}{
-	"library": map[string]interface{}{
-		"name":    "analytics-go",
-		"version": Version,
-	},
-}
```
achille-roussel: I made this unexported and moved to `context.go`.

message.go
```diff
@@ -27,3 +27,13 @@ func makeMessageId(id string, def string) string {
...
+
+// This structure represents objects sent to the /v1/batch endpoint. We don't
+// export this type because it's only meant to be used internally to send groups
+// of messages in one API call.
+type batch struct {
+	MessageId string        `json:"messageId"`
+	SentAt    string        `json:"sentAt"`
+	Messages  []interface{} `json:"batch"`
+	Context   Context       `json:"context"`
+}
```
achille-roussel: I simplified the `batch` type, it's used only once internally and therefore didn't need to satisfy the `Message` interface.

time.go
```diff
@@ -30,3 +30,10 @@ func makeTime(t time.Time, def time.Time) time.Time {
 	}
 	return t
 }
+
+// Returns a string representation of the time value passed as first argument,
+// unless it's a zero-value, in that case the default value passed as second
+// argument is used instead.
+func makeTimestamp(t time.Time, def time.Time) string {
+	return formatTime(makeTime(t, def))
+}
```
achille-roussel: This didn't need to be part of this PR but I noticed that in one place I had forgotten a call to `makeTime`, to avoid this mistake in the future I added this helper function.

context.go
```diff
@@ -0,0 +1,162 @@
...
+// This type provides the representation of the `context.os` object as defined
+// in https://segment.com/docs/spec/common/#context
+type OSInfo struct {
+	Name    string `json:"name,omitempty"`
+	Version string `json:"version,omitempty"`
+}
+
+// This type provides the representation of the `context.page` object as
+// defined in https://segment.com/docs/spec/common/#context
+type PageInfo struct {
```
achille-roussel: Here I've suffixed the context sub-types with `-Info` because there's already a type named `Page` (for the page message).
I noticed that the Java library was naming the messages with a `-Message` suffix (like `PageMessage`), if we were following this naming convention we could get rid of the `-Info` suffix. However I feel like `Track`, `Page`, `Identify`, etc... are used way more often than the context sub-types so I feel better about keeping the shorter names for what's most common.
Please let me know if you have a different point of view.

context.go
```diff
@@ -0,0 +1,162 @@
...
+	Longitude float64 `json:"longitude,omitempty"`
+	Speed     float64 `json:"speed,omitempty"`
+}
+
+// This type provides the representation of the `context.network` object as
+// defined in https://segment.com/docs/spec/common/#context
+type NetworkInfo struct {
+	Bluetooth bool   `json:"bluetooth,omitempty"`
+	Cellular  bool   `json:"cellular,omitempty"`
+	WIFI      bool   `json:"wifi,omitempty"`
```
achille-roussel: Here I'm not sure about the data type, my understanding is that it should be booleans but the documentation isn't very clear about it: https://segment.com/docs/spec/common/#context

analytics.go
f2prateek: I believe the reason this was exported so that users can attach their own default context to all the calls.

context.go
f2prateek: Yup this should be a boolean.

context.go
f2prateek: Yeah I'm not particularly attached to any convention, we use `Payload` in the mobile libraries. I think the short names SGTM.

analytics.go
f2prateek: (We don't leave to need it exported but just for reference).

achille-roussel: Oh interesting, you mean by setting values of the `DefaultContext` variable then it gets picked up automatically?

Global state is usually not a good practice, how about making this part of the client configuration when it's instantiated?

f2prateek: Agreed, plus I'm not comfortable giving users the option to override the library information (we should be source of truth of that). it'd be neat if we could add something even more powerful like the Java library and provide an interceptor mechanism (https://github.com/segmentio/analytics-java/blob/master/analytics/src/main/java/com/segment/analytics/MessageInterceptor.java#L14-L19). This would let users implement coming logic in a single place rather than at every site where they make a tracking call.

f2prateek:
LGTM!

I'm not a big fan of having the `Extras` field, but I don't see a way around it.

achille-roussel:
Cool, thanks for reviewing!

PR #78, opened 2016-04-24
Config

### LINKED ISSUE #43: Calling close before sending anything blocks forever

### LINKED ISSUE #51: Make analytics.Client an interface
@f2prateek 

What do you think about making the `analytics.Client` an interface, from the current PRs it looks like we're moving toward having all fields being unexported, which means there's no much need for the actual type to be exposed.

The interface can look like this:
```go
type Client interface {
    io.Closer

    Alias(*Alias) error

    Page(*Page) error

    Group(*Group) error

    Identify(*Identify) error

    Track(*Track) error
}
```
This way we could also have a `type MockClient struct{}` which implements this interface and can be used in test code for example.

Let me know what you think about it.

### LINKED ISSUE #68: len(msgs) >= c.Size
This would prevent the buffer never being emptied if someone sets to size to 0:

```go
// https://github.com/segmentio/analytics-go/blob/master/analytics.go#L316
for {
		select {
		case msg := <-c.msgs:
			c.verbose("buffer (%d/%d) %v", len(msgs), c.Size, msg)
			msgs = append(msgs, msg)
			if len(msgs) == c.Size {
				c.verbose("exceeded %d messages – flushing", c.Size)
				c.send(msgs)
				msgs = nil
```

achille-roussel:
@f2prateek 
@calvinfo 

Fixes #43 
Fixes #51 
Fixes #68 

Hey guys, these changes fix the issue we had due to the configuration being exposed within the `Client` type.
Client objects are now exposed as a simple interface and don't leak their internal structure anymore.


config.go
```diff
@@ -0,0 +1,142 @@
...
+	// timer triggers.
+	Interval time.Duration
+
+	// The HTTP transport used by the client, this allows an application to
+	// redefine how requests are being sent at the HTTP level (for example,
+	// to change the connection pooling policy).
+	// If none is specified the client uses `http.DefaultTransport`.
+	Transport http.Transport
+
+	// The logger used by the client to output info or error messages when that
```
paddie: 'when' should be removed, yeah?

config.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if c.Transport == nil {
+		c.Transport = http.DefaultTransport
+	}
+
+	if c.Logger == nil {
+		c.Logger = newDefaultLogger()
+	}
+
+	if c.BatchSize == 0 {
```
paddie: Should batchSize == 0 be an invalid size? If someone wanted the client to send individual events in realtime as they occurred (which is not in your/segment's interest) - but someone might expect this to be a valid value.

config.go
```diff
@@ -0,0 +1,142 @@
...
+	// A function called by the client to get the current time, `time.Now` is
+	// used by default.
+	Now func() time.Time
+
+	// The retry policy used by the client to resend requests that have failed.
+	// The function is called with how many times the operation has been retried
+	// and is expected to return how long the client should wait before trying
+	// again.
+	// If not set the client will fallback to use a default retry policy.
+	RetryAfter func(int) time.Duration
```
paddie: Clever!

config.go
paddie: Maybe just document our way out of this one.

config.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(c.Endpoint) == 0 {
+		c.Endpoint = DefaultEndpoint
+	}
+
+	if c.Interval == 0 {
+		c.Interval = DefaultInterval
+	}
+
+	if c.Transport == nil {
+		c.Transport = http.DefaultTransport
```
f2prateek: I think that instead of saving a transport - we should create the `http.Client` once here and save a reference to the client.

This way we aren't creating a new http client for each http request (https://github.com/segmentio/analytics-go/blob/v3.0/analytics.go#L134).

config.go
f2prateek: Sending events in realtime would have a batch size of 1 (which is definitely acceptable value, e.g. during development).

config.go
```diff
@@ -0,0 +1,151 @@
...
+	// If none is specified the client uses a standard logger that outputs to
+	// `os.Stderr`.
+	Logger Logger
+
+	// The maximum number of messages that will be sent in one API call.
+	// Messages will be sent when they've been queued up to the maximum batch
+	// size or when the flushing interval timer triggers.
+	// Note that the API will still enfore a 500KB limit on each HTTP request
+	// which is independant from the number of embedded messages.
+	BatchSize int
```
f2prateek: Unsigned?

config.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// When set to true the client will send more frequent and detailed messages
+	// to its logger.
+	Verbose bool
+
+	// A function called by the client to generate unique message identifiers.
+	// The client uses a UUID generator if none is provided.
+	UID func() string
```
f2prateek: I don't know if this should be exposed.

config.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// A function called by the client to get the current time, `time.Now` is
+	// used by default.
+	Now func() time.Time
```
f2prateek: Not sure if this should be exposed either.

config.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// The retry policy used by the client to resend requests that have failed.
+	// The function is called with how many times the operation has been retried
+	// and is expected to return how long the client should wait before trying
+	// again.
+	// If not set the client will fallback to use a default retry policy.
+	RetryAfter func(int) time.Duration
+
+	// The default context set on each message sent by the client.
+	Context Context
```
f2prateek: I think it's clearer to name this `DefaultContext`

config.go
achille-roussel: Yeah, 1 works. With a batch size of 1 as soon as one message gets queued it will get sent since it reaches the batch size (there's already a test to confirm that).

config.go
achille-roussel: That's a good point, the client we use internally should be persisted across requests, I'll fix.

config.go
achille-roussel: I tend to default to using `int` everywhere usually even for stuff that can be negative just to make it easy to integrate with most of the Go code out there that uses `int` and rarely `uint` (think functions like `len` which return `int`, you suddenly need a type conversion if you want to inject it here).

The main issue is that a `uint` cast from an integer value that is negative is totally valid but likely produce a value that wasn't the writer's intent. If we let negative values be set here at least we'll catch them at runtime and return an error instead of creating a client with a batch size of a few billion messages.

Discussion on that: https://www.reddit.com/r/golang/comments/2q5vdu/int_vs_uint/

config.go
achille-roussel: Do you mean it should be an unexported field of the `Config` type or just exposed in a different way?
We have to be able to set it somehow so unit tests can mock it to produce predictable values.

config.go
achille-roussel: OK

config.go
achille-roussel: That gets me thinking, should we also have a hardcoded maximum value for the batch size? it doesn't make much sense to allow batches of more than 1000 for example since it's likely gonna be rejected by the API?

f2prateek: I think that's ok (1000). A user might want to set the size to something arbitrarily large because they want to control exactly when a message gets flushed. Especially once we do this (https://github.com/segmentio/analytics-go/issues/81) we'll split the payload anyway.

config.go
f2prateek: Correct, I meant it should unexported so we can use it for unit tests.

config.go
f2prateek: I wonder if we should give users more control and supply the error that prompts a retry. Not entirely sure if that would be useful.

achille-roussel: I think that for the 99.99% cases application don't need to worry about retries and that's kind of why they'd use this library in the first place, they need something that already handle failure cases.
Not to mention that the error is supposed to be a one-in-a-billion times event, especially for server-to-server communication where connections are all wired and high quality (which this library is designed for).

I feel like ultimately we may want to figure out a way to report messages that could not be sent at all so the application can figure out what they need to do with it but I doubt this will have

config.go
achille-roussel: OK I'll change it then.

config.go
achille-roussel: OK

config.go
f2prateek: so we do this in the Java library — https://github.com/segmentio/analytics-java/blob/master/analytics/src/main/java/com/segment/analytics/Callback.java

The use case being some customers (I think Simple was the one that originally requested this) have their own queueing infrastructure for analytics events. Specially they had a RabbitMQ queue, that their worker pulled from, used our library to upload it, and the callback let them ack the message to indicate it had been processed.

achille-roussel: I see, I think we can introduce this concept outside of this PR if you don't mind, just to not make too many changes at once.

f2prateek: Makes sense!

analytics.go
```diff
@@ -3,85 +3,123 @@ package analytics
...
-		uid:       uid,
-	}
+type client struct {
+	Config
+	key string
+
+	// This channel is where the `Enqueue` method writes messages so they can be
+	// picked up and pushed by the backend goroutine taking care of applying the
+	// batching rules.
+	msgs chan interface{}
```
f2prateek: Would it  be better to make this `chan message` rather than `interface{}`, and offload serialization to the grouting too?

analytics.go
```diff
@@ -3,85 +3,123 @@ package analytics
...
+
+	m := msg.serializable(c.uid(), c.now())
+
+	defer func() {
+		// When the `msgs` channel is closed writing to it will trigger a panic.
+		// To avoid letting the panic propagate to the caller we recover from it
+		// and instead report that the client has been closed and shouldn't be
+		// used anymore.
+		if recover() != nil {
+			err = io.EOF
```
f2prateek: should be have our own error (`var ErrClosed = errors.New("client closed")`) instead of `EOF`?

config.go
f2prateek: @achille-roussel since we're allowing negative values, we should change the `if c.BatchSize == 0 {` check to check `<= 0` I think.

achille-roussel: The way I was seeing it is `validate` checks for invalid values, once `makeConfig` is called the configuration object has already been validated so the `BatchSize < 0` case can never trigger.

analytics.go
achille-roussel: I thought about it, I don't think it should be a change blocking this PR (the historical data type for this field has been `chan interface{}`) but I'll consider this option once I tackle #81

analytics.go
achille-roussel: I can make this change, I usually use `EOF` to express that an I/O object has reached a state where it can't process anything anymore, plus it reuses an error that a developer is already familiar with.
I can make the change if you feel like it's more appropriate.
Should we also return `ErrClosed` in the `Close` method when it's called more than once? (we currently use `io.EOF` there as well).

f2prateek: Yeah unsure - from the docs:

```
EOF is the error returned by Read when no more input is available. Functions should return EOF only to signal a graceful end of input. If the EOF occurs unexpectedly in a structured data stream, the appropriate error is either ErrUnexpectedEOF or some other error giving more detail.
```

To me ErrClosedPipe sounds better, but down for either (EOF, this, or ErrClosedPipe)
```
ErrClosedPipe is the error used for read or write operations on a closed pipe.
```

f2prateek:
LGTM, just a few nits inline but nothing blocking!

analytics.go
achille-roussel: OK, I'll add an ErrClosed then, it does make more sense.

PR #84, opened 2016-04-25
predict batch size + simplify

### LINKED ISSUE #21: Checking for error in track function
Is there a way to check for an error in the `track` function?

### LINKED ISSUE #81: Predict the size of the batched request
@f2prateek 
@calvinfo 

I noticed that we don't check for the size of the serialized JSON before sending a request and only rely on the batch size (number of in-flight messages).
Is there a historical reason for that or did we just not spend the time implementing it because limiting the message count works well enough?

### LINKED ISSUE #83: Add Callback interface to notify the application of message send success or failure
Port what is done in the Java library: https://github.com/segmentio/analytics-java/blob/master/analytics/src/main/java/com/segment/analytics/Callback.java

achille-roussel:
@f2prateek 

Fixes #21 
Fixes #81 
Fixes #83 

Hey man, this PR adds the logic to predict what the batch size is gonna be and force sending the request if that size is exceeded.

The `chan interface{}` field of the client is also replaced with a `chan Message` so the serialization is done in the background.

There's a new `Callback` interface that can be defined on clients (the way it's done in the Java library) to be notified when message sends have succeeded or failed.

I also pulled in the logic from #50 to make requests asynchronous and not block the goroutine calling `Enqueue`.

Finally I got rid of the *serializable* concept in favor of exposing the `Type` field in the message structures and a type switch in `Client.Enqueue` to set the default values.

Please check it out and let me know if you need anything to be changed.


f2prateek:
can you send these as separate PRs?

config.go
```diff
@@ -48,7 +52,7 @@ type Config struct {
 	Verbose bool
 
 	// The default context set on each message sent by the client.
-	DefaultContext Context
+	DefaultContext *Context
```
achille-roussel: This was changed to a pointer for consistency with other types that now embed a `*Context` as well. Using a pointer instead of the value is necessary to have the `omitempty` tag work (zero-value structs are not considered *empty*).

achille-roussel:
Sure

PR #86, opened 2016-04-25
limit message batches by size of serialized payload

achille-roussel:
@f2prateek 

Changes related to limiting message batches to 500KB as described in #84

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -3,6 +3,7 @@ package analytics
 import (
 	"io"
 	"io/ioutil"
+	"sync"
 
 	"bytes"
 	"encoding/json"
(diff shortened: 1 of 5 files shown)


f2prateek:
This feels wasteful that  we marshal a message twice - once for length, and once for an upload.

achille-roussel:
I don't think this is what the code is doing, the marshaled JSON is cached in the `message` object which implemented `json.Marshaler` and just returns the value it has cached instead of re-marshaling.

See: https://github.com/segmentio/analytics-go/pull/86/files#diff-b95c50d14ed7b4bb1cfecdc4e498d4dcR67

f2prateek:
Ah I see, nice! Seeing `json.Marshal` twice confused me.

achille-roussel:
:) `json.Marshaler` is a sexy pretty sexy way of injecting stuff in the JSON encoder.

Let me know if that's good to go.

f2prateek:
LGTM!

PR #90, opened 2016-04-25
add Traits type

achille-roussel:
@f2prateek 

This adds a `Traits` type in the style of what's done in the Java library, basically a map with a couple of helper methods to set the common fields.

PS: I think we should add `Integration` and `Properties` types as well, even if there are no helper methods yet for these, at least that let us plan for upcoming release and we won't have to make API breaking changes, what do you think about it?


achille-roussel:
Moved to #91

PR #91, opened 2016-04-25
add Traits type

achille-roussel:
@f2prateek

This adds a Traits type in the style of what's done in the Java library, basically a map with a couple of helper methods to set the common fields.

PS: I think we should add Integration and Properties types as well, even if there are no helper methods yet for these, at least that let us plan for upcoming release and we won't have to make API breaking changes, what do you think about it?

### DIFF
--- a/group.go
+++ b/group.go
@@ -15,7 +15,7 @@ type Group struct {
 	GroupId      string                 `json:"groupId"`
 	Timestamp    time.Time              `json:"timestamp,omitempty"`
 	Context      *Context               `json:"context,omitempty"`
-	Traits       map[string]interface{} `json:"traits,omitempty"`
+	Traits       Traits                 `json:"traits,omitempty"`
 	Integrations map[string]interface{} `json:"integrations,omitempty"`
 }
 
--- a/traits.go
+++ b/traits.go
@@ -22,7 +22,7 @@ func NewTraits() Traits {
 	return make(Traits, 10)
 }
 
-func (t Traits) SetSetAddress(address string) Traits {
+func (t Traits) SetAddress(address string) Traits {
 	return t.Set("address", address)
 }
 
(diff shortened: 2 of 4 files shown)


traits.go
```diff
@@ -0,0 +1,92 @@
...
+//			.Set("Role", "Jedi"),
+//	}
+//
+type Traits map[string]interface{}
+
+func NewTraits() Traits {
+	return make(Traits, 10)
+}
+
+func (t Traits) SetSetAddress(address string) Traits {
```
f2prateek: I think this should be SetAddress.

traits.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (t Traits) SetFirstName(firstName string) Traits {
+	return t.Set("firstName", firstName)
+}
+
+func (t Traits) SetGender(gender string) Traits {
+	return t.Set("gender", gender)
+}
+
+func (t Traits) SetID(id string) Traits {
```
f2prateek: I think this is deprecated and should be removed (people should send userid instead).

group.go
```diff
@@ -15,7 +15,7 @@ type Group struct {
 	GroupId      string                 `json:"groupId"`
 	Timestamp    time.Time              `json:"timestamp,omitempty"`
 	Context      *Context               `json:"context,omitempty"`
-	Traits       map[string]interface{} `json:"traits,omitempty"`
+	Traits       Traits                 `json:"traits,omitempty"`
```
f2prateek: Could this be left `map[string]interface{}`? That way users can opt in tot type safety with the `Traits` if they want, but ignore if they'd prefer map.

f2prateek:
LGTM, there aren't that many spec'd properties so could be ok to skip them. Here's what we have in Android (mostly commerce ones).

traits.go
achille-roussel: OK I'll change, the specs don't mention that yet.

group.go
achille-roussel: No, the Go type system doesn't allow it, `type Traits map[string]interface{}` is not assignable to `map[string]interface{}`.

Although this syntax is still supported:
```go
group := Group{
    ...
    Traits: analytics.Traits{
        "name": "foo",
    }
}
```
So it's just a matter of typing `analytics.Traits` instead of map[string]interface{}, and if the code was using `map[string]interface{}` already it can be easily converted as well:
```go
traits := map[string]interface{}{
    "name": "foo",
}

group := Group{
    ...
    Traits: analytics.Traits(traits),
}
```

PR #92, opened 2016-04-26
Properties

achille-roussel:
@f2prateek 

Follow up of #91, adds a Properties type that mirrors what is done in the Android library.


achille-roussel:
@f2prateek do you mind checking this really quick?

f2prateek:
Sweet, LGTM!

PR #93, opened 2016-04-26
add Integrations type

achille-roussel:
@f2prateek 

Follow up of #91, add the `Integrations` type which defines helper functions to turn on/off integrations.


f2prateek:
Can you document that set with non false will enable?

achille-roussel:
Sure

f2prateek:
LGTM otherwise.

PR #97, opened 2016-04-26
limit the number of goroutines

achille-roussel:
@f2prateek 

Following up on our conversation, this should limit the number of goroutines spawned by the client and block if too many are in use already.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -127,6 +127,12 @@ type Client struct {
 	now      func() time.Time
 	once     sync.Once
 	wg       sync.WaitGroup
+
+	// These synchronization primitives are used to control how many goroutines
+	// are spawned by the client for uploads.
+	upmtx   sync.Mutex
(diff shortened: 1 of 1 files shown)


f2prateek:
Neat, haven't used sync.Cond before. I think we can simplify by using a semaphore directly though https://github.com/f2prateek/semaphore if we want to simplify.

LGTM!

achille-roussel:
Yep, a semaphore primitive would be great to have in Go... I don't know if Go optimizes allocation of `make(chan struct, 1000)` to not allocate 1KB+ of memory for the semaphore, if it does it's great, if not I'd rather use `sync.Cond` + counter, which seems like a more efficient way to manage the semaphore.

The Channel has the advantage of being injectable in a `select`, I'll probably do that in 3.0.0 to wait for `c.quit` as well.

PR #98, opened 2016-04-26
add executor to limit the number of goroutines

achille-roussel:
@f2prateek 

Following up on our conversation and the changes we've made for 2.1.1, this adds the executor concept to limit the number of goroutines being executed.

This is mostly just a cleanup / testable version of what's in 2.1.1, but it calls the failure callback if the executor is full instead of blocking.

Please take a look and let me know if you're OK with the changes!

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -170,9 +170,10 @@ func (c *client) Close() (err error) {
 }
 
 // Asychronously send a batched requests.
-func (c *client) sendAsync(msgs []message, wg *sync.WaitGroup) {
+func (c *client) sendAsync(msgs []message, wg *sync.WaitGroup, ex *executor) {
 	wg.Add(1)
-	go func() {
+
+	if !ex.do(func() {
 		defer wg.Done()
 		defer func() {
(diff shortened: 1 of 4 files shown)


f2prateek:
Oooh nice, I'm a bit unsure if need a heavy handed executor approach here though - the advantage of doing this in java was that users could supply their own to control the execution, but this doesn't really apply to Go IMO since we don't give users this ability. I think just a simple counter will do, no?

achille-roussel:
The main reason I pulled it in its own object if for testing purposes, that way we can test this part of the logic in isolation and verify it works instead of having it baked into the client implementation, plus I feel like it makes the client implementation easier to read since all the sync logic is abstracted.

The minimum we need is a mutex+counter, although I feel like this piece of code could be easily reused internally based on what we choose to do for #99.

I'll hold on until we've got that figured out before making changes to this PR if you don't mind.

f2prateek:
Ok testing internally sounds good, feel free to merge!

PR #100, opened 2016-04-27
100% test coverage

achille-roussel:
@f2prateek 

This PR adds 100% test coverage for the analytics package:
```
$ go test -cover     
PASS
coverage: 100.0% of statements
ok  	github.com/segmentio/analytics-go	0.202s
```
Going through this process allowed me to find a couple bugs, so it was definitely worth it (one bad key set on Properties, and failure callbacks not being called due to errors being dropped).

This should give us a good base to evolve the package in the future, as well as more confidence when time comes to make this release.

Please take a look and let me know if you see something that should be changed.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -258,7 +258,7 @@ func (c *client) upload(b []byte) error {
 func (c *client) report(res *http.Response) (err error) {
 	var body []byte
 
-	if res.StatusCode < 400 {
+	if res.StatusCode < 300 {
 		c.debugf("response %s", res.Status)
 		return
 	}
--- a/analytics.go
+++ b/analytics.go
@@ -1,6 +1,7 @@
 package analytics
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"sync"
@@ -203,7 +204,7 @@ func (c *client) send(msgs []message) {
 	})
 
 	if err != nil {
-		c.errorf("marshalling mesages - %s", err)
+		c.errorf("marshalling messages - %s", err)
 		c.notifyFailure(msgs, err)
 		return
 	}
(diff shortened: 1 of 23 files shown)


analytics.go
```diff
@@ -250,25 +251,25 @@ func (c *client) upload(b []byte) error {
...
-	return nil
+	return c.report(res)
 }
 
 // Report on response body.
-func (c *client) report(res *http.Response) {
+func (c *client) report(res *http.Response) (err error) {
+	var body []byte
+
 	if res.StatusCode < 400 {
```
achille-roussel: Do you know why this condition was expressed this way? It feels to me that anything but a 200 should be considered an error (the [doc](https://segment.com/docs/libraries/http/#errors) says the API always returns 200).

f2prateek: True, though I think we should still stick to normal HTTP semantics (2xx/3xx OK and 4xx, 5xx not ok). We've talked internally about changing the 200 OK to 202 Accepted and being too aggressive with the error codes might bite us later.

f2prateek:
LGTM!

analytics.go
achille-roussel: OK.

My only concern is the http client should be handling 3xx automatically already, so getting one at this level kind of feels like a bug / error and we likely shouldn't accept this as a valid response.

How about changing it to `res.StatusCode < 300`?

f2prateek: sure, i don't mid either way!

f2prateek: OH I also realized our Api doesn't return 200 always (https://github.com/segmentio/tracking-api/blob/9b18d72c74944369243da16a12e255b959640b19/server/server.go#L65-L71), it will return  a 400 if your json is invalid.

achille-roussel: Yep, hopefully the library don't ever generate invalid JSON 🙏

PR #101, opened 2016-04-27
add HTTP timeout

achille-roussel:
@f2prateek 

Adds a 10 second timeout on HTTP requests as a safety measure, we can figure out a better logic to set this value in the future.


f2prateek:
LGTM

PR #102, opened 2016-04-27
define callback policy

### LINKED ISSUE #99: Async Callback
@f2prateek 

I think we should define the policy on how the success and failure callbacks are called. Right now they're triggered from different places in the code, which could end up creating dead locks if the application has locked a resource that its callback would also try to acquire, or just race conditions if the app makes assumptions on what goroutine notifies its callbacks.

Here are few options we can explore:

- Have an internal queue + goroutine that's only in charge of the calling the application's callback so all function calls are made from a single goroutine

- Spawn a new goroutine for every batch of messages we want to notify the application of.

- Use an internal queue + executor so we dispatch callbacks concurrently with a limit on how many goroutines will be created

Whatever option we choose I'll make sure to document it in the `Callback` type.
Let me know if you have other ideas or what sounds the best approach to you.

achille-roussel:
@f2prateek 

Fixes #99

### DIFF
--- a/message.go
+++ b/message.go
@@ -7,6 +7,14 @@ import (
 
 // Values implementing this interface are used by analytics clients to notify
 // the application when a message send succeeded or failed.
+//
+// Callback methods are called by a client's internal goroutines, there are no
+// guarantees on which goroutine will trigger the callbacks, they could be made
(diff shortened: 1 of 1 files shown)


f2prateek:
LGTM!

PR #103, opened 2016-04-27
add code coverage artifacts

achille-roussel:
@f2prateek 

This PR adds detailed code coverage reports in the Circle-CI artifacts so we can keep a trace of the package's health.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ _testmain.go
 *~
 \#*
 .\#*
+
+# Artifacts
+tmp/*
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+ifndef CIRCLE_ARTIFACTS
+CIRCLE_ARTIFACTS=tmp
+endif
+
 vet:
 	@go vet ./...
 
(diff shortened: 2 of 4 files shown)


achille-roussel:
@f2prateek alright I finally got that this to build and all tests to run... the issue was tests were not being ran in the GOPATH which was breaking generating the coverage report.

The coverage reports now get added to the artifacts, you can check out the one for the last build https://circle-artifacts.com/gh/segmentio/analytics-go/264/artifacts/0/tmp/circle-artifacts.eu46e6Y/cover.html

Please let me know if you want to see anything changed.

f2prateek:
Cool,  LGTM! How does this work - as in it just stores the reports in Circle? I think it'd be neat to report code coverage to a tool like stat hat on every merge to master so we can track the number over time.

Also should consider moving to Travis CI maybe - their build matrix support is nice so we can run CI against multiple targets https://github.com/f2prateek/train/blob/master/.travis.yml#L4-L6

achille-roussel:
Yep Circle-CI stores the artifact along with the build. We can definitely integrate with a code coverage tool to get this tracked over time, the ones I've used so far did not provide the ability to store the coverage reports tho.

Travis has upsides (like what you mentioned, their integration with Go is also much better) but they don't host the build artifacts so we won't get this for free.
The html reports that Go generate are pretty good to get visibility into what is covered and what not, having this part of the CI is really convenient, we can probably figure out a work around tho.

Thanks for reviewing!

f2prateek:
Ah interesting. I think I can run the code coverage html reports on my own computer since I doubt I'd go on Circle to see what I missed. Not sure I'd miss it if we moved to Travis. IMO getting just a number is good enough from the CI.

achille-roussel:
Yeah it can definitely be ran locally, but having it as an online resource is always handy, for tracking purposes and for code reviews for example.

PR #104, opened 2016-04-27
fix http timeout tests

achille-roussel:
@f2prateek 

There were issues with Go 1.5 (which Circle-CI is using apparently, despite them saying they're on 1.6), the test HTTP transport did not implement `CancelRequest` which made the HTTP client refuse to send requests after the timeout was set.

### DIFF
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -28,6 +28,8 @@ type roundTripperFunc func(*http.Request) (*http.Response, error)
 
 func (f roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) { return f(r) }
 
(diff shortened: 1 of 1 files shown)


achille-roussel:
This was part of #103, I'll merge it since it was reviewed already.

f2prateek:
We should use conditional compilation (https://golang.org/pkg/go/build/) to add this depending on the Go version.

achille-roussel:
Do you mean we should do something like not setting the HTTP client timeout if the Go version is <= 1.5 and the `http.RoundTripper` doesn't support cancellation?

f2prateek:
yup!

PR #105, opened 2016-04-28
fix go 1.5 compat on cancelable requests

achille-roussel:
@f2prateek 

I couldn't figure out how to do conditional compilation based on go versions so I went for a runtime check.
I initially wanted to use `runtime.Version()` but it may return the commit number instead of a version number if the compiler is not in a release version so it's not very reliable.

Instead I've relied on the changes that were made to the `net/http` package between 1.5 and 1.6 regarding request cancellation.

Please take a look and let me know if you're satisfied with the changes.


f2prateek:
Cool, LGTM! I also sent a PR with #106 with conditional compilation (either approach works - up to you).

achille-roussel:
I like to use a compile time check better but I couldn't find a clear example of how to do that well in the docs.

PR #106, opened 2016-04-28
Set default http timeout conditionally.

f2prateek:
Use conditional compilation to set default timeout on http client.
For go 1.6+ we always the http timeout.
For other older versions, we set the timeout only if the transport
supports it.

cc @achille-roussel 

Closes #105

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -85,16 +85,23 @@ func NewWithConfig(writeKey string, config Config) (cli Client, err error) {
 		msgs:     make(chan Message, 100),
 		quit:     make(chan struct{}),
 		shutdown: make(chan struct{}),
-		http: http.Client{
-			Transport: config.Transport,
-			Timeout:   10 * time.Second,
-		},
+		http:     makeHttpClient(config.Transport),
 	}
 
 	go c.loop()
 
 	cli = c
 	return
+}
+
+func makeHttpClient(transport http.RoundTripper) http.Client {
+	httpClient := http.Client{
+		Transport: transport,
+	}
+	if supportsTimeout(transport) {
+		httpClient.Timeout = 10 * time.Second
+	}
+	return httpClient
 }
 
 func (c *client) Enqueue(msg Message) (err error) {
--- a/timeout_15.go
+++ b/timeout_15.go
@@ -1,0 +1,16 @@
+// +build !go1.6
+
+package analytics
+
+import "net/http"
+
+// http clients on versions of go before 1.6 only support timeout if the
+// transport implements the `CancelRequest` method.
+func supportsTimeout(transport http.RoundTripper) bool {
+	_, ok := transport.(requestCanceler)
+	return ok
+}
+
+type requestCanceler interface {
+	CancelRequest(*http.Request)
+}
--- a/timeout_16.go
+++ b/timeout_16.go
@@ -1,0 +1,10 @@
+// +build go1.6
+
+package analytics
+
+import "net/http"
+
+// http clients on versions of go after 1.6 always support timeout.
+func supportsTimeout(transport http.RoundTripper) bool {
+	return true
+}


http_15.go
```diff
@@ -0,0 +1,16 @@
+// +build !go1.6
```
achille-roussel: Sexy, where did you find this in the docs? I couldn't figure it out. 

Also is this working with non-release versions (like tip for example)?

http_16.go
```diff
@@ -0,0 +1,10 @@
+// +build go1.6
```
achille-roussel: How is this gonna work with Go 1.7? Do we have to update the build tags every new go release?

analytics.go
```diff
@@ -57,7 +57,7 @@ type client struct {
 
 	// This HTTP client is used to send requests to the backend, it uses the
 	// HTTP transport provided in the configuration.
-	http http.Client
+	http *http.Client
 }
```
achille-roussel: What's the reason you changed this to a pointer?
The http.Client object's lifetime is bound to the analytics.client's, it feels like it's giving more work to the GC for no reason.

http_15.go
f2prateek: https://golang.org/pkg/go/build/

not sure, as long as tip reports a version higher than go1.6 we should be good.

http_16.go
f2prateek: Nope, this flag is true for all versions 1.6 or higher.

analytics.go
f2prateek: I might be missing something here, but my understanding was that the http client should always be used as a pointer?

I'll revert this and we discuss it separately.

f2prateek:
Removed pointer update.

f2prateek:
realize the file name is probably better off named `timeout_....go`

analytics.go
f2prateek: So first thing was because the default http client is a pointer (https://golang.org/pkg/net/http/#pkg-variables).

Secondly, reading https://golang.org/doc/faq#Pointers "Each variable in Go exists as long as there are references to it.", so not sure why using a pointer causes more work for the GC?

achille-roussel: There's no hard rule on whether a program should use a pointer or a value, the language automatically takes the address for methods with pointer receivers ;)

f2prateek: Interesting, I think I originally just copied what I saw here https://github.com/google/go-github/blob/master/github/github.go#L74 and have been using that since!

analytics.go
```diff
@@ -97,6 +94,16 @@ func NewWithConfig(writeKey string, config Config) (cli Client, err error) {
 	return
 }
 
+func newHttpClient(transport http.RoundTripper) http.Client {
```
achille-roussel: Go conventions are to name `make-` functions returning values (`new-` is for functions returning pointers).

achille-roussel:
The changes look good to me, thanks for taking the time to make this 👍

f2prateek:
Updated new to make and rename http_ to timeout_

analytics.go
achille-roussel: There's one memory allocation done for the `analytics.client` object, if the `http.Client` is a value, the memory for this object will be part of the `analytics.client`. It it's a pointer Go needs to make another memory allocation for the `http.Client` and the `analytics.client` then references this other memory area, the GC then now has to manage 2 pointers intstead of one. I agree it's not a big deal in this case, it's just a good practice.

I feel like using pointers vs values allows to express ownership. When an object is gonna be shared, using a pointer makes sense because different part of the application a referencing the same object. When there's only one owner, you can get away with just a value and get better memory management.
Obviously in Go you can choose to get the address of pretty much anything and start creating shared ownership of a value, even if it wasn't declared as a pointer initially, the only downside is it could result in preventing a larger block of memory from being released (think if the object is part of a slice or struct), but usually you don't need to care about this and can just fallback to using a value.

Using pointers also have runtime small runtime costs besides putting more work on the GC because the language does nil-checking every time a pointe is dereferenced, with a value the compiler knows that the value can't ever be nil so it can optimize this away.

Other use-cases for pointers are to describe that the value is allowed to be missing (like with the `omitempty` flag in the json package), but in our case we don't ever want the http client to not exist either.

achille-roussel:
Awesome :+1:

morenoh149:
What's blocking this?

achille-roussel:
Nothing :)

PR #107, opened 2016-06-22
New CleanHyphen format

stinkyfingers:


f2prateek:
Is there a compelling reason to use this new format?

PR #110, opened 2016-08-11
Provide an interface to the Client.

cam-stitt:
There is great value in having an exported interface to the client. We are currently using analytics-go within our application, but wanted to have our client mocked out for testing purposes.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -107,8 +107,8 @@ type Alias struct {
 	Message
 }
 
-// Messager is an interface for the segment client.
-type Messager interface {
+// Messenger is an interface for the segment client.
+type Messenger interface {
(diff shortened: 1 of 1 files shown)


f2prateek:
For the current version, I think this is easy enough to write in your own code base.

In the next version v3, the client will only be available as an interface https://github.com/segmentio/analytics-go/blob/v3.0/analytics.go#L18-L39.

cam-stitt:
Yeah, we already have it in or code base, thought it was a small enough
change. Look forward to version 3.

f2prateek:
feel free to try out version 3, it's ready for production and very well tested, would to get your more users trying it out so we can incorporate any feedback!

PR #112, opened 2017-05-19
Adding "DevinceInfo.AdvertisingID" in order to send data to AppsFlyer integration

corgrath:
Adding "DevinceInfo.AdvertisingID" since it is required in order to send
Server-to-Server events to AppsFlyer according to Segment documentation:
https://segment.com/docs/integrations/appsflyer/#server-side-tracking


f2prateek:
Thanks

PR #113, opened 2017-07-25
Resource leak in loop

wyattjoh:
Fixed resource leak on loop ticker.

As per the docs on the ticker: https://golang.org/pkg/time/#NewTicker

> ... Stop the ticker to release associated resources.

This just adds that.


wyattjoh:
Totally missed that you close that when you call quit. My bad!

PR #114, opened 2017-09-23
Library improvenments

m1ome:
Added interface Logger;
Added options support;
Added NewWithOptions function;
Refactoring errors;


f2prateek:
Thanks - however we have a v3 (https://github.com/segmentio/analytics-go/tree/v3.0), which includes most of these changes.

PR #122, opened 2018-03-14
Add CLI to send messages to Segment.

f2prateek:
This adds a CLI that conforms to our testing harness contract. https://paper.dropbox.com/doc/Libraries-End-to-End-Testing-Harness-dnotyIVIGYR7lnO1LcZtM#:uid=457845578223014078216812&h2=Contract. This allows the testing harness to test the Go library.

Since the harness doesn't have any fixtures using the anonymousId, context and integrations top level fields, they haven't been added to the CLI yet either.

PR #123, opened 2018-03-14
Migrate to Circle CI v2

f2prateek:
When working on https://github.com/segmentio/analytics-go/pull/122, I realized the Circle configuration was pretty old and didn't support the latest go version.

I put this together based on https://github.com/segmentio/conf/blob/master/.circleci/config.yml and https://discuss.circleci.com/t/how-to-test-multiple-versions-by-triggering-jobs-with-a-shell-function/11305/15.

Currently we test against the following go versions; latest, 1.10, 1.9 and 1.8. I'm not sure what the Go version distribution is like, but 1.8 was released over a year ago, so this seemed reasonable. Eventually maybe we should have better understanding of what our customers are using so we can make smarter decisions.

Ref: LIB-196


f2prateek:
Blocking #122

Makefile
```diff
@@ -11,8 +14,7 @@ build:
 test:
 	@mkdir -p ${CIRCLE_ARTIFACTS}
 	@go test -race -coverprofile=${CIRCLE_ARTIFACTS}/cover.out .
-	@go test -race -cover ./vendor/...
```
f2prateek: I removed this line - it wasn't working with the new setup, and didn't make sense to me to test the vendored libraries anyway...

achille-roussel: Sounds good to me 👍

.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
+# Inspired by https://discuss.circleci.com/t/how-to-test-multiple-versions-by-triggering-jobs-with-a-shell-function/11305/15.
+version: 2
+jobs:
+  build-golatest:
+    docker:
+      - image: circleci/golang
+    working_directory: /go/src/github.com/segmentio/analytics-go
+    steps:
+      - checkout
+      - run: make get vet test
```
achille-roussel: What do you think about adding a `ci` rule to the Makefile? Then you just do `make ci`, if it changes there's no need to update the circleci config.

f2prateek: Done!

PR #124, opened 2018-03-16
Add End to End tests.

f2prateek:
This integrates the testing harness with the Go library. The harness invokes the cli package with fixtures and verifies that the library was able to send those messages to a webhook connected to the project succesfully.

https://paper.dropbox.com/doc/Libraries-End-to-End-Testing-Harness-dnotyIVIGYR7lnO1LcZtM

2 main additions:

1. A new CLI that conforms to our testing harness contract. https://paper.dropbox.com/doc/Libraries-End-to-End-Testing-Harness-dnotyIVIGYR7lnO1LcZtM#:uid=457845578223014078216812&h2=Contract. This allows the testing harness to test the Go library.

2. Run the harness on CI.

### DIFF
--- a/cmd/cli/main.go
+++ b/cmd/cli/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/segmentio/analytics-go"
+	"github.com/segmentio/conf"
+)
+
+func main() {
+	var config struct {
+		WriteKey   string `conf:"writeKey"   help:"The Segment Write Key of the project to send data to"`
+		Type       string `conf:"type"       help:"The type of the message to send"`
+		UserID     string `conf:"userId"     help:"Unique identifier for the user"`
+		GroupID    string `conf:"groupId"    help:"Unique identifier for the group"`
+		Traits     string `conf:"traits"     help:"Metadata associated with the user"`
+		Event      string `conf:"event"      help:"Name of the track event"`
+		Properties string `conf:"properties" help:"Metadata associated with an event, page or screen call"`
+		Name       string `conf:"name"       help:"Name of the page/screen"`
+	}
+	conf.Load(&config)
+
+	callback := callback(make(chan error, 1))
+
+	client, err := analytics.NewWithConfig(config.WriteKey, analytics.Config{
+		BatchSize: 1,
+		Callback:  callback,
+	})
+	if err != nil {
+		fmt.Println("could not initialize analytics client", err)
+		os.Exit(1)
+	}
+
+	switch config.Type {
+	case "track":
+		client.Enqueue(analytics.Track{
+			UserId:     config.UserID,
+			Event:      config.Event,
+			Properties: parseJSON(config.Properties),
+		})
+	case "identify":
+		client.Enqueue(analytics.Identify{
+			UserId: config.UserID,
+			Traits: parseJSON(config.Traits),
+		})
+	case "group":
+		client.Enqueue(analytics.Group{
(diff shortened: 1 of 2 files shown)


cmd/cli/main.go
```diff
@@ -0,0 +1,96 @@
...
+	conf.Load(&config)
+
+	callback := newCallback()
+
+	client, err := analytics.NewWithConfig(config.WriteKey, analytics.Config{
+		BatchSize: 1,
+		Callback:  callback,
+	})
+	if err != nil {
+		panic(err)
```
achille-roussel: Is it worth panicking here? it'll give a stack trace that isn't very interesting to the end user.

How about one of these options?
```go
fmt.Println(err)
os.Exit(1)
```
```go
log.Fatal(err)
```

cmd/cli/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+func main() {
+	var config struct {
+		WriteKey   string `conf:"writeKey"   help:"The Segment Write Key of the project to send data to"`
+		Type       string `conf:"type"       help:"The type of the message to send"`
+		UserID     string `conf:"userId"     help:"Unique identifier for the user"`
+		GroupID    string `conf:"groupId"    help:"Unique identifier for the group"`
+		Traits     string `conf:"traits"     help:"Metadata associated with the user"`
+		Event      string `conf:"event"      help:"Name of the track event"`
+		Properties string `conf:"properties" help:"Metadata associated with an event, page or screen call"`
```
achille-roussel: The `conf` package supports map types, why not use that so it gets loaded automatically instead of getting a string and parsing it?

cmd/cli/main.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// callback implements the analytics.Callback interface. It is used by the CLI
+// to wait for events to be uploaded before exiting.
+type callback struct {
+	success chan struct{}
+}
+
+func (c *callback) Failure(_ analytics.Message, err error) {
+	panic(err)
```
achille-roussel: This will crash the application with a stack trace in an internal goroutine.

What do you think about changing the code like this:
```go
type callback chan error

func (c callback) Failure(_ analytics.Message, err error) {
    c <- err
}

func (c callback) Success(_ analytics.Message) {
    c <- nil
}
```
```go
// in main
if err := <-callback; err != nil {
    fmt.Println(err)
}
```

cmd/cli/main.go
f2prateek: Didn't know it did that, updating!

cmd/cli/main.go
f2prateek: Didn't pay too much attention to error messages since this is just for testing atm, but will update!

cmd/cli/main.go
f2prateek: will update!

f2prateek: Done!

cmd/cli/main.go
f2prateek: Done!

cmd/cli/main.go
f2prateek: Turns out the objconv pacakge doesn't handle the `interface{}` in `map[string]interface{}` well. unfortunately we need this so we can exercise the library with a wider range of fixtures, so we're sticking with manual parsing for now. I improved the error message in case of a JSON parsing failure though!

achille-roussel: Actually objconv can deal with this just fine, it’s conf that has trouble with it ;)

:+1:

PR #126, opened 2018-04-19
Fix #125 AppInfo.Build type

### LINKED ISSUE #125: AppInfo failed for Unmarshal event type track
Hi,

I use `analytics-go` for unmarshal json received by the webhook.

> json: cannot unmarshal number into Go struct field AppInfo.build of type string.

Could you normalized the type before sending it to the webhook or use the type `json.Number` for `AppInfo.Build` ?

see: https://golang.org/pkg/encoding/json/#Number

Thx

euskadi31:
Fix #125


f2prateek:
See 
https://github.com/segmentio/analytics-go/issues/125#issuecomment-463318549

PR #137, opened 2018-10-02
Add changelog for 3.0.1

systemizer:


History.md
```diff
@@ -1,3 +1,10 @@
+v3.0.1 / 2018-10-02
+===================
+* Migrate from Circle V1 format to Circle V2
```
f2prateek: I think we can skip these (or mark them differently) in the future as it's an internal change (doesn't affect consumers of the library).

PR #138, opened 2018-12-12
fail when enqueuing invalid message

pwfcurry:
`client.Enqueue(&msg)` silently fails - I've modified the method to return an error instead.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -1,6 +1,7 @@
 package analytics
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


pwfcurry:
Ideally this should be handled, but my go reflection-foo is not strong.

f2prateek:
@achille-roussel any thoughts on if we should return an error here, or  maybe automatically unwrap the pointer instead?

pwfcurry:
Can we get this merged? Or at least discuss options. Thanks!

f2prateek:
My sense is that this should be handled by the library (instead of an error) - if the compiler accepts it, we should too.

f2prateek:
Gonna close this PR as I don't think this is right behaviour we want to accept, and instead unwrap the pointer as mentioned in 
https://github.com/segmentio/analytics-go/pull/138#issuecomment-454918697

f2prateek:
Mostly solved by #146 now, but logging might be a good idea for custom types. If you'd like to re-open this, please add a test case!

Issue #140, opened 2019-02-13
v3: maxMessageBytes differs from docs


jeffreylo:
It seems that the `maxMessageBytes` in the package is set to 15000 bytes (15KB), but the docs suggest that it should actually be 32KB:

>There is a maximum of 500KB per batch request and 32KB per call.

— https://segment.com/docs/sources/server/go/#batching

Is this intentional?

PR #141, opened 2019-02-13
message: update maxMessageBytes to 32KB

jeffreylo:
This patch raises the maxMessageBytes allowed per call to 32KB as stated
in documentation for V3.

> There is a maximum of 500KB per batch request and 32KB per call.

— [segment.com/docs/sources/server/go/#batching](https://segment.com/docs/sources/server/go/#batching)

Closes https://github.com/segmentio/analytics-go/issues/140.


f2prateek:
LGTM!

Issue #145, opened 2019-03-05
Event tracking: passing a pointer to `Enqueue` results in silent dropped events


philipithomas:
We just had a problem with our code while switching to the V3 api, and it resulted in dropped events with no other symptoms, errors, or logs. It turns out the problem was that we were passing a reference to the track. This passed linting and tests. It did not return an error. Its logs made it look like the event was successfully uploaded. 

This was broken:

```
err = client.Enqueue(&analytics.Track{

        // Event name is action + target, e.g. create_payment

        Event:  "test-event",

        UserId: userID,

    })

    if err != nil {

        panic(err)

    }
```

This one works:

```
err = client.Enqueue(analytics.Track{
        // Event name is action + target, e.g. create_payment

        Event:  "test-event",

        UserId: userID,

    })

    if err != nil {

        panic(err)

    }
```

My request: Can you add some checks to prevent this from silently happening in the future?

f2prateek:
This has been fixed with #146

PR #146, opened 2019-04-02
Handle pointer types in Enqueue()

andreiko:
Trying to fix this issue https://github.com/segmentio/analytics-go/issues/145 in a way that @f2prateek suggested here https://github.com/segmentio/analytics-go/pull/138/

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -104,7 +104,45 @@ func makeHttpClient(transport http.RoundTripper) http.Client {
 	return httpClient
 }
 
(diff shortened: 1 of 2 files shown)


lieut-data:
@andreiko, will there be a new release of this package containing this fix? We just ran into the same issue, and while we've fixed our code, I can imagine others encountering this same pain point when they upgrade.

andreiko:
Hello @lieut-data, I just released version `v3.1.0` containing this change.

PR #148, opened 2019-04-04
return error if a custom type is enqueued

f2prateek:
Because of the way we construct messages with type assertion, custom types that satisfy the `Message` interface cannot be used.

Normally these would be silently dropped, but this returns an error so that failure is more obvious.


f2prateek:
alternatively, we can panic here, as after #146, there's never a case where this code path can be triggered by consumers of the library, as the `Message` interface cannot be implemented externally.

Screen Shot 2019-04-03 at 6 57 52 PM

PR #149, opened 2019-04-14
GoDeps: fix missing dependencies, and switch to `vendor` directory

thaJeztah:
Some dependencies were missing; added the missing dependencies, updated jehiah/go-strftime (which was a bit out of date), and switched to use a `vendor` directory.

Note that Godeps is no longer maintained, so should likely be replaced for another vendoring tool, but I kept this for now.


thaJeztah:
@williamgrosset PTAL

f2prateek:
The master branch contains code for the v2 version which has been deprecated and is no longer maintained.

I don't think this issue applies to v3, but you can send a PR on the v3.0 branch.

If you'd like to continue using the changes on v2, I recommend using a fork.

PR #151, opened 2019-04-14
fix missing vendors, re-vendor with LK4D4/vndr

thaJeztah:
I didn't see a configuration file for a vendoring tool, so used https://github.com/LK4D4/vndr for now; alternatively the whole vendor directory could be removed


f2prateek:
We didn't want to vendor dependencies for the `cli` and `examples` package, as they're not meant for consumers of the library (e.g. the mold package). Not sure if there's a better way to handle that.

thaJeztah:
if no specific versions of the dependencies are needed, and people are mainly expected to use this repo as a dependency, perhaps remove the vendor directory altogether?

thaJeztah:
Closing as there's a go.mod now

PR #153, opened 2019-06-13
Update context.go

sumankarki:
Added source field in context struct


abraithwaite:
The `source` field in `context` is not part of our spec.

https://segment.com/docs/spec/common/#context

Please use the `Extra` struct field to access `source` if needed.

If that doesn't work for you, please elaborate in the issue you opened regarding this.

PR #156, opened 2019-06-21
Expose the Message interface Validate method

deankarn:
By exposing the Message interfaces validate() method it allows the
analytics types to be used externally in other code and not confined to
the analytics-go library.

### DIFF
--- a/message.go
+++ b/message.go
@@ -37,7 +37,7 @@ type Message interface {
 	// nil if the message is valid, or an error describing what went wrong.
 	Validate() error
 
-	// internal is an unexposed interface function to ensure only Message types defined within this package are usable.
+	// internal is an unexposed interface function to ensure only types defined within this package can satisfy the Message interface. Invoking this method will panic.
 	internal()
 }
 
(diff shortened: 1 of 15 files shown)


track.go
```diff
@@ -19,7 +21,11 @@ type Track struct {
 	Integrations Integrations `json:"integrations,omitempty"`
 }
 
-func (msg Track) validate() error {
+func (msg Track) internal() {
+	panic("implement me")
```
f2prateek: maybe we should do this for all of the types?

message.go
```diff
@@ -33,9 +33,12 @@ type Callback interface {
 // and therefore can be passed to the analytics.Client.Send method.
 type Message interface {
 
-	// Validates the internal structure of the message, the method must return
+	// Validate validates the internal structure of the message, the method must return
 	// nil if the message is valid, or an error describing what went wrong.
-	validate() error
+	Validate() error
+
+	// internal is an unexposed interface function to ensure only Message types defined within this package are usable.
```
f2prateek:
```suggestion
	// internal is an unexposed interface function to ensure only types defined within this package can satisfy the Message interface. Invoking this method will panic.
```

PR #160, opened 2020-05-27
Remove typed events

tysonmote:


tysonmote:
(Just to be clear, this is just a spike to see what it might look like to remove type validation for some internal use cases. I have no intention of ever seeing this merged to HEAD. 😄)

PR #163, opened 2020-06-10
Add Message implementation for a generic message type

kalamay:
This adds a new `Event` generic message type that implements the `Message` interface for a `map[string]interface{}`. This validates the `"type"` key with a required value of `"alias"`, `"group"`, `"identify"`, `"page"`, `"screen"`, or `"track"`, and it follows the event type specific validation rules. While all messages created by `analytics-go` are supported by the Tracking API, there are some messages that the Tracking API supports that this library does not. It is still preferable to use the more restrictive type definitions where possible, but there are some cases where we need the flexibility to send messages that do not fit this stricter definition. The purpose of this addition is to allow sending these messages while still keeping the validation rules shared within this project.

We need to use this for two of our internal services. I've also proposed PR #169 which removes the `internal()` requirement of the `Message` interface. One of these approaches is needed for `webhook-functions-consumer` and `platform-api`.

### DIFF
--- a/generic_message.go
+++ b/generic_message.go
@@ -2,7 +2,7 @@ package analytics
 
 var _ Message = GenericMessage(nil)
 
-// This type represents any event type sent in a track call as described in
+// This type represents any event type sent to the Tracking API as described in
 // https://segment.com/docs/libraries/http/
 type GenericMessage map[string]interface{}
 
--- a/generic_message.go
+++ b/generic_message.go
@@ -10,41 +10,52 @@ func (msg GenericMessage) internal() {
 	panic(unimplementedError)
 }
 
+func (msg GenericMessage) string(field string) string {
+	val, _ := msg[field].(string)
+	return val
+}
+
 func (msg GenericMessage) Validate() error {
 	if typ, ok := msg["type"].(string); ok {
 		switch typ {
 		case "alias":
-			m := Alias{Type: "alias"}
-			m.UserId, _ = msg["userId"].(string)
-			m.PreviousId, _ = msg["previousId"].(string)
-			return m.Validate()
+			return Alias{
+				Type:       "alias",
+				UserId:     msg.string("userId"),
+				PreviousId: msg.string("previousId"),
+			}.Validate()
 		case "group":
-			m := Group{Type: "group"}
-			m.UserId, _ = msg["userId"].(string)
-			m.AnonymousId, _ = msg["anonymousId"].(string)
-			m.GroupId, _ = msg["groupId"].(string)
-			return m.Validate()
+			return Group{
+				Type:        "group",
+				UserId:      msg.string("userId"),
+				AnonymousId: msg.string("anonymousId"),
+				GroupId:     msg.string("groupId"),
+			}.Validate()
 		case "identify":
-			m := Identify{Type: "identify"}
-			m.UserId, _ = msg["userId"].(string)
-			m.AnonymousId, _ = msg["anonymousId"].(string)
-			return m.Validate()
+			return Identify{
+				Type:        "identify",
+				UserId:      msg.string("userId"),
+				AnonymousId: msg.string("anonymousId"),
+			}.Validate()
 		case "page":
-			m := Page{Type: "page"}
-			m.UserId, _ = msg["userId"].(string)
-			m.AnonymousId, _ = msg["anonymousId"].(string)
-			return m.Validate()
+			return Page{
+				Type:        "page",
+				UserId:      msg.string("userId"),
+				AnonymousId: msg.string("anonymousId"),
+			}.Validate()
 		case "screen":
-			m := Screen{Type: "screen"}
-			m.UserId, _ = msg["userId"].(string)
-			m.AnonymousId, _ = msg["anonymousId"].(string)
-			return m.Validate()
+			return Screen{
+				Type:        "screen",
+				UserId:      msg.string("userId"),
+				AnonymousId: msg.string("anonymousId"),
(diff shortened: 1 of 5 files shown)


generic_message.go
```diff
@@ -0,0 +1,55 @@
+package analytics
+
+var _ Message = GenericMessage(nil)
+
+// This type represents any event type sent in a track call as described in
+// https://segment.com/docs/libraries/http/
+type GenericMessage map[string]interface{}
```
achille-roussel: I wonder if a name like `Map` would have more leverage: it would communicate that it's a `map`, is shorter, and more descriptive than "Generic" (which is also an overloaded term in programming languages).

generic_message.go
```diff
@@ -0,0 +1,55 @@
...
+		case "screen":
+			m := Screen{Type: "screen"}
+			m.UserId, _ = msg["userId"].(string)
+			m.AnonymousId, _ = msg["anonymousId"].(string)
+			return m.Validate()
+		case "track":
+			m := Track{Type: "track"}
+			m.UserId, _ = msg["userId"].(string)
+			m.AnonymousId, _ = msg["anonymousId"].(string)
+			m.Event, _ = msg["event"].(string)
```
achille-roussel: I wonder if all these type asserts make it worth it to have a method on the type of `msg` to lookup strings to improve readability, for example:
```go
m.Event = msg.string("event")
```
Then we could also write:
```go
m := Track{
    Type:  "track",
    Event: msg.string("event"),
    ...
}
```

generic_message.go
```diff
@@ -0,0 +1,55 @@
+package analytics
+
+var _ Message = GenericMessage(nil)
+
+// This type represents any event type sent in a track call as described in
```
achille-roussel: This comment seems inaccurate, it seems that the values may be more than track calls only.

generic_message.go
```diff
@@ -0,0 +1,55 @@
...
+// https://segment.com/docs/libraries/http/
+type GenericMessage map[string]interface{}
+
+func (msg GenericMessage) internal() {
+	panic(unimplementedError)
+}
```
achille-roussel: This internal method being part of the interface definition of `Message` is the reason you couldn't declare this type in your own program I believe.

Maybe it's an indicator that we should get rid of it and let users extend the package with their own types wherever they need?

generic_message.go
kalamay: Yeah, I think that would clean it up a bit.

generic_message.go
kalamay: Ah, the intention there was to express any event sent through the Tracking API, but I completely fumbled that description!

generic_message.go
kalamay: Yep, that is the reason, and removing `internal()` is what I was going to propose initially. Maybe that was the right direction? The rationale for _not_ going that way though is that we'd get at least some minimal "standardized" validation logic provided by the library rather than distributing that around to each time a flexible event type is needed. Though it seems this is the first time?

generic_message.go
kalamay: Ah fair point, I'm surely in agreement to remove the "Generic" part of the name, but I'd lean towards something descriptive. Maybe `MapMessage`? I'll keep thinking on the name though.

generic_message.go
achille-roussel: Maybe we can solve this too, via an exported validation function, or some other mechanism (e.g. if a custom type claims it's a track, verify that it contains the required fields).

Since `Message` has never been implemented by external types (due to `internal()`), we can still change the interface without breaking programs to gain the features we want.

generic_message.go
achille-roussel: Looking at prior art, the `mgo` package for example just uses `type M map[string]interface{}`. `logrus` uses `type Fields map[string]interface{}`.

Generally, Go names shy away from binding to an interface they're intended to implement. Keeping short and descriptive is often a successful approach.

kalamay: I am aware of Go's conventions here, but we're not trying to say that it is a map in the general sense. It is a message that uses a `map` to carry its payload. Calling it a `Map` is short, but not descriptive in my opinion. In the case of logging, you send it fields to log, so calling it `Fields` describes the purpose of the map, not so much that it _is_ a map. The difference is in the semantics of _what it is_ vs. _what is its purpose_, and I tend to find conveying purpose to be more important.

I do agree that `GenericMessage` doesn't quite name this well, but it seems we don't have a good generic-use term. I had started by just calling it `Event`, but in the docs, that term is more associated with `track` events specifically.

generic_message.go
kalamay: It's right that we can safely do this now, but it would mean opening up the `Message` interface from here on out (for this version at least), and this seems a bit more of a considered change. Is there some further need that sways one way or the other? I find it hard to advocate for a particular direction given a sample size of 1. 😃 

The way I'm breaking this down in my mind is that we have two fundamental needs:

1. A way to send messages that aren't defined by the existing types.
2. Some way to validate that a less restrictive message meets some minimal structural requirements.

We could open up the `Message` interface, which takes care of the first point. But it doesn't enforce the second point. We could provide a hypothetical `Validate(map[string]interface{})` function to do this, but there wouldn't be any way to ensure that the message had been validated by it. If we think that is reasonable, it would fulfill our needs. I just have some initial reservations about the opening the interface.

achille-roussel: It does seem to go beyond the intent of the PR, so probably better to postpone and keep this discussion as content for the time we choose to tackle this.

PR #164, opened 2020-07-06
Use go mod for v3.0 branch

hblanks:
As above! Use go mod for this branch.

But also:

* Fix circle CI build (it was failing due to two `jobs` stanzas)
* Update Readme so build badge points to our branch.

After this merges, we should be able to also update the godoc badge to point to the v3 docs. (Right now, the v3 docs tag has `+incompatible`. . . I gather because we don't have a `go.mod` on the branch yet.)


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/segmentio/analytics-go
+
+go 1.14
+
+require (
+	github.com/segmentio/backo-go v0.0.0-20200129164019-23eae7c10bd3
+	github.com/segmentio/conf v1.2.0
+	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c
```
achille-roussel, line 8: Maybe a change for a subsequent PR, we could replace this dependency with `github.com/google/uuid` which is a preferred UUID package.

go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/segmentio/analytics-go
+
+go 1.14
+
+require (
+	github.com/segmentio/backo-go v0.0.0-20200129164019-23eae7c10bd3
+	github.com/segmentio/conf v1.2.0
```
achille-roussel, line 7: Having this package imported also brings lots of other dependencies, I wonder if it would be worth revisiting the code to remove this dependency.

achille-roussel (approved):

go.mod
hblanks: I'll take a look!

go.mod
hblanks: Noted, and can do some time this week.

go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/segmentio/analytics-go
```
thaJeztah, line 1: @achille-roussel  @hblanks This looks incorrect. Go modules that are `v2` or above MUST use the major version as part of the module name. `go modules` accepts `incompatible` versions (`v3.0.1+incompatible`) if the code does not have a `go.mod`, but if there's a `go.mod` present, it will reject the module. See https://blog.golang.org/v2-go-modules

Given that this project already released `v3.xx` versions _without_ a `go.mod`, adding a `go.mod` therefore _requires_ the major version to be updated ([SemVer](https://semver.org), which in this case means;

- rename the module to `github.com/segmentio/analytics-go/v4`
- release as `v4.0.0`

I would suggest reverting this change for the v3 branch, and starting a v4 branch for a v4.0.0 release.

hblanks: Hi, and sorry for the delay. I'm not sure either @achille-roussel or I are maintainers for this repo, but I can try to help here best I can.

Your note was a helpful summary on `go mod` behavior, but I'm not sure how to map such best practices onto this repo, where we have a separate branch for v3 instead of a separate directory running on the master branch. I'd agree that ideally we'd cut a new version, but it seems like we'd prefer to do that only if there are either new features or a user need for it. I definitely submitted this PR so that I could use this repo with `go mod`. It does seem to continue to be working for people since then.

Are you having trouble building software using the v3 branch of this repo? If so, please let me know with some details - again, I'm not the maintainer but I'd help out much as I'm able to.

thaJeztah: > I'm not sure how to map such best practices onto this repo, where we have a separate branch for v3 instead of a separate directory running on the master branch

Yeah, their "best practice" on using subdirectories is a bit silly. The Go maintainers have the opinion that releases are always tagged from master/main, and not to use release branches. Generally, it should be ok to use release branches, but som older dependency management tools (`go deps`, e.g.) won't work without some trickery at least.

>  I'd agree that ideally we'd cut a new version, but it seems like we'd prefer to do that only if there are either new features or a user need for it.

Understood; usually that would be ok, but in this case it's a bit unfortunate that a release _was_ done, but introduced a breaking change.

SemVer dictates that the major version is not to indicate "new features", but for "breaking changes" (adding a go.mod in this case is). An easy rule of thumb for SemVer is;

    v<breaking>.<features and enhancements>.<bug fixes>

- If a release only contains bug fixes, increment the last digit
- If a release contains new features, increment the middle digit
- If a release contains a breaking change, increment the first digit.

>  It does seem to continue to be working for people since then.

Let me write up a quick example to show where / when it's causing issues.

thaJeztah: Ah, perhaps it works better if I actually post the example (LOL, got pulled into a chat on our Slack)

You can observe the issue if you try to use the latest (`v3.2.0`) version as a module:

Create a project, and initialize it as a module:

```bash
$ mkdir myproject && cd myproject
$ go mod init myproject
```

Add code that uses this project:

```bash
cat > main.go <<EOF
package main

import (
    "fmt"
    "github.com/segmentio/analytics-go"
)

func main() {
    fmt.Println(analytics.Version)
}
EOF
```

Try to add the `v3.2.0` version. Go modules will refuse to use that version, because the module name is incompatible:

```bash
$ go get github.com/segmentio/analytics-go@v3.2.0

go get: github.com/segmentio/analytics-go@v3.2.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

As a workaround, it's possible to use the git commit corresponding with `v3.2.0` instead. This "tricks" go modules into using it as a non-released `v1.x.x` module, and it will generate a pseudo version for it, based on the last tag based of master/main;

```
$ go get github.com/segmentio/analytics-go@0566e489c7b96d42123ee2733f267a667e84d8ff
go: downloading github.com/segmentio/analytics-go v1.2.0
go: downloading github.com/segmentio/analytics-go v1.2.1-0.20201110202747-0566e489c7b9
go: downloading github.com/segmentio/backo-go v0.0.0-20200129164019-23eae7c10bd3
go: downloading github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c
go get: added github.com/segmentio/analytics-go v1.2.1-0.20201110202747-0566e489c7b9
```

Notice how `go mod` generated a pseudo version (`v1.2.1-0.20201110202747-0566e489c7b9`)


Tidy the modules and check what's in `go.mod` now;

```bash
$ go mod tidy && cat go.mod

module myproject

go 1.16

require (
	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
	github.com/kr/pretty v0.3.0 // indirect
	github.com/segmentio/analytics-go v1.2.1-0.20201110202747-0566e489c7b9
)
```

Code is able to run (looks like the `Version` const hasn't been updated since `v3.0.0`, so doesn't show the correct version)

```bash
$ go run -mod=mod main.go
3.0.0
```

Because of the incompatible version, users that want to get the latest version, don't actually get that version; go modules will "accept" incompatible versions for projects that don't have a `go.mod`, but once added, it will simply ignore releases that are incompatible. As a result; if you try to get the latest; `go` will pick the latest release that didn't have a `go.mod`:

```bash
go get github.com/segmentio/analytics-go@latest
go: downloading github.com/segmentio/analytics-go v3.1.0+incompatible
go get: upgraded github.com/segmentio/analytics-go v1.2.1-0.20201110202747-0566e489c7b9 => v3.1.0+incompatible

$ go mod tidy && cat go.mod

module myproject

go 1.16

require (
	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
	github.com/kr/pretty v0.3.0 // indirect
	github.com/segmentio/analytics-go v3.1.0+incompatible
	github.com/segmentio/backo-go v0.0.0-20200129164019-23eae7c10bd3 // indirect
	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
)
```

PR #169, opened 2020-11-10
Remove message internal-only restriction and add ValidateFields

kalamay:
This change removes the `internal()` method requirement for the `Message` interface. This is added so that only internal types may implement the `Message` interface. However, we have a need in two of our services to implement a custom message type with less restrictive value requirements. The `Validate()` method of the various event types in this library enforce encoding fields with stricter values then are required by the Tracking API. For some of our internal services that invoke the Tracking API, we need to match the flexibility it provides. By removing the `internal()` requirement, we can implement our own `Message` that matches these expectations.

Additionally, this defines a `ValidateFields(map[string]interface{}) error` function. This verifies that the input contains an accepted event type specifier, and then passes minimally required fields into the type-specific validation. This can then be used to validate and/or construct a `Message` object in the other services that need to maintain full Tracking API compatibility (`webhook-functions-consumer`, `platform-api`, etc.)

A prior attempt at this was to define another, more flexible `Message` implementation in PR #163. We need to move ahead with either of these two PRs.

Additionally, this fixes some code that needs to convert an `int` into a `string` when adding a header.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -1,0 +1,56 @@
+package analytics
+
+func getString(msg map[string]interface{}, field string) string {
+	val, _ := msg[field].(string)
+	return val
+}
+
+func ValidateFields(msg map[string]interface{}) error {
+	if typ, ok := msg["type"].(string); ok {
+		switch typ {
+		case "alias":
+			return Alias{
+				Type:       "alias",
+				UserId:     getString(msg, "userId"),
+				PreviousId: getString(msg, "previousId"),
+			}.Validate()
+		case "group":
+			return Group{
+				Type:        "group",
+				UserId:      getString(msg, "userId"),
+				AnonymousId: getString(msg, "anonymousId"),
+				GroupId:     getString(msg, "groupId"),
+			}.Validate()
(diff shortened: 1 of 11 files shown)


validate.go
```diff
@@ -0,0 +1,56 @@
+package analytics
+
+func getString(msg map[string]interface{}, field string) string {
+	val, _ := msg[field].(string)
+	return val
+}
+
+func ValidateFields(msg map[string]interface{}) error {
```
achille-roussel: Enforcing a `map[string]interface{}` may be a bit too opinionated on the context in which this function would be used.

Have you considered other approaches?

tysonmote (approved):

validate.go
kalamay: We could go back to the interface that requests the underlying field values, ie:

```
type FieldGetter interface {
    GetField(field string) (interface{}, bool)
}
```

And use that as the type to `ValidateFields`. 

What other use cases do you have in mind?

dk1027 (approved):

PR #172, opened 2021-02-09
Add Context.Direct

### LINKED ISSUE #171: Missing Context.Direct
According to https://segment.com/docs/connections/sources/catalog/libraries/server/http-api/#collecting-ip-address : 

> When sending a HTTP call from a user’s device, you can collect the IP address by setting context.direct to true.

There seems to be no `Context.Direct` in analytics-go.

We're trying to get telemetry from a CLI app, but would rather not manually figure out and send the IP for geolocation in the `IP` field.

kke:
Fixes #171 

Adds the `direct` field to context, which can be used when using the analytics-go package in a client application to avoid having to set the IP address on the client side.

### DIFF
--- a/context.go
+++ b/context.go
@@ -20,6 +20,7 @@ type Context struct {
 	Referrer  ReferrerInfo `json:"referrer,omitempty"`
 	Screen    ScreenInfo   `json:"screen,omitempty"`
 	IP        net.IP       `json:"ip,omitempty"`
+	Direct    bool         `json:"direct,omitempty"`
 	Locale    string       `json:"locale,omitempty"`
 	Timezone  string       `json:"timezone,omitempty"`
 	UserAgent string       `json:"userAgent,omitempty"`


kke:
Tested working, the Track event now gets an IP in the segment debugger view when Direct is true.

Before:

```json
{
  "context": {
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    },
}
```

After:

```json
{
  "context": {
    "direct": true,
    "ip": "8***.2***.6***.1***",
    "library": {
      "name": "analytics-go",
      "version": "3.0.0"
    },
  }
}
```

jasmingacic (approved):

kke:
@f2prateek @sanscontext Is anyone maintaining this package anymore?

prayansh (approved):
@kke thx for the PR. apologies for the delay on the feedback. changes LGTM, so gonna go ahead and merge it

PR #179, opened 2021-11-30
wait for requests to finish on flush

kushmansingh:
This helps guarantee that all requests are made when the quit signal is received. The previous `wg.Done()` did nothing since it was deferred.


swtch1:
This change is needed.  Can someone provide a review or merge?

PR #186, opened 2022-08-23
fix typo in Product string json tag

ajanssens:


TonyZZX:
I wonder, would it break any backward compatibility?

PR #188, opened 2022-10-04
Update docs

converge:
`go get https://github.com/segmentio/analytics-go` give the following error:
```
go: malformed module path "https:/github.com/segmentio/analytics-go": invalid char ':'
```

the example code could handle the error.


converge:
`go get https://github.com/segmentio/analytics-go` give the following error:
```
go: malformed module path "https:/github.com/segmentio/analytics-go": invalid char ':'
```

the example code could handle the error.

PR #189, opened 2023-01-14
Add group id to context

nickdelja:
Add group id to context so the track events can related to both a user and distinct id and group.


kalamay (approved):

kalamay:
Overall, this looks good to me. I'd recommend that any other project that updates to this version needs to verify they aren't currently using the `Extra` map to find the `groupId` because this new field will now take precedence.

PR #206, opened 2026-05-21
Dual-path retry: exponential backoff + rate-limit handling

MichaelGHSeg:
## Summary

Replaces the fixed 10-attempt retry loop with a structured dual-path retry system that separates rate-limiting (429 + Retry-After) from general transient failures.

- **429 + Retry-After header**: sleep for the specified duration (capped at 300s), does NOT consume the retry budget. Bounded by `MaxRateLimitDuration` (default 12h).
- **Other retryable errors** (5xx, 408, 410, 460): counted exponential backoff (base 500ms, 2×, cap 60s). Bounded by `MaxRetries` (default 10) and `MaxTotalBackoffDuration` (default 12h).
- **Non-retryable errors** (4xx except the above, plus 501/505/511): discard immediately.
- Removes `select { case <-c.quit }` from retry sleep paths — `Close()` now waits for in-flight retries to finish naturally via `wg.Wait()` instead of killing them mid-flight.
- Adds `X-Retry-Count` header on retry attempts (omitted on first attempt).
- Wraps body-read errors in `httpError` preserving the original status code's retryability (fixes a bug where a non-retryable 4xx with a broken response body would be incorrectly retried).
- New config fields: `MaxRetries`, `MaxTotalBackoffDuration`, `MaxRateLimitDuration`.
- New sentinel errors: `ErrBackoffBudgetExceeded`, `ErrRateLimitBudgetExceeded`.
- E2E: enables `retry` test suite; wires `maxRetries` from e2e-cli input config.

## Test plan

- [ ] `go test -race ./...` passes
- [ ] E2E `basic,retry` suites pass (48/48)


analytics.go
```diff
@@ -260,28 +269,74 @@ func (c *client) send(msgs []message) {
...
+
+		backoffAttempts++
+		if backoffAttempts > c.MaxRetries {
+			c.errorf("%d messages dropped after %d attempts", len(msgs), totalAttempts)
+			c.notifyFailure(msgs, lastErr)
+			return
+		}
+
+		time.Sleep(c.RetryAfter(backoffAttempts - 1))
+	}
```
abueide, line 381: this function is pretty big, I'd split it up to make it a bit more readable and debuggable, try to make the flow of figuring out what action is being taken on each loop easy to understand, right now its really difficult to reason about where the function will return, sleep or continue.

MichaelGHSeg: Split it up into helpers

didiergarcia (approved):

